PR #44, opened 2015-04-09
quick fixes for compatibility and search problem

dowcet:
Line 714: The idea here is to only catch keys if they are a `:BIBTEX-KEY:` property or at the start of an org-mode headline. Someone familiar with using regexps in org-mode could do this much better.

Lines 716-718: I don't see `outline-hide-other` and `outline-show-subtree` defined in Emacs 24.3.1. I'm not sure if my change will behave differently, but it seems OK so far.


dowcet:
I'm realizing that this definitely needs to be a `re-search-forward`, because there may be more then once space after `:BIBTEX-KEY:`.

Issue #40, opened 2015-02-19
Request for helm-bibtex-notes-file


jagrg:
Instead of storing my notes in a separate directory, which I think is what helm-bibtex proposes, I like to keep my notes in one org file, which looks something like this:

    ** Author (2013) Title
    :PROPERTIES:
    :Custom_ID: bibtex-key
    :END:

So I was wondering if you could add "helm-bibtex-notes-file", which would be the equivalent of "helm-bibtex-notes-path". But instead of searching for the bibtex-key.org file, it would search for the bibtex-key in the file specified by the user, for example:

    (setq helm-bibtex-notes-file "~/.emacs.d/annotation.org")

tmalsburg:
I considered storing all notes in one file when I first wrote helm-bibtex.  The reason why I decided against it is that in my experience Emacs can be really slow with bigger org files. So I thought that the one-file solution may not be scaleable.

Are there any true benefits of having all notes in one file or is this just a matter of personal preference?

If a one-file solution were to be added, I think we would simply recycle the variable `helm-bibtex-notes-path`. If it points to a directory, make one file per publication.  If it points to an org-file, store the notes in that file.  For that file, I would probably prefer to have a structure like the following:

    * DoeEtAl2013
    Here are my notes on Jon Doe's recent paper: ...

where `DoeEtAl2013` is the BibTeX key.  While this is perhaps less human-readable, it has the benefit that it avoids the property drawer, which seem to slow down Emacs a lot.  (I believe org-contacts is so unbearably slow because it heavily relies on properties.)

I'm just thinking aloud here.  I'm not planning to work on this and I may not even accept a pull-request.  At the moment, it seems to me that the one-file approach is neither better or worse than the approach using multiple files and I don't see why we need two solutions for storing notes.

jag2:
The reason for using one file would be to extend the flexibility of helm-bibtex. People have different approaches. Some work better than others. Personally, I've always used one file and never had problems with speed, even when having to retrieve from drawers.

Some of the benefits of using one file is to be able to (1) export all or some of the notes to html, pdf etc.; (2) search with helm-swoop or occur/org-occur; (3) sort and move notes to different subtree categories. Another advantage is that the note heading and metadata are added automatically using the information provided in the bib file, which avoids having to name files in a separate directory.

tmalsburg:
[...]

PR #67, opened 2015-10-19
Add option for supporting the interleave package

cpaulik:
I've implemented support for the [interleave](https://github.com/rudolfochrist/interleave) package for myself and thought it might be interesting for others.

If you don't want to support this since users can easily change the note taking function then feel free to close.


tmalsburg:
I didn't know about the interleave package.  Looks really useful.  There is a branch (note-files, see #40) which includes support for note file templates.  I would probably include interleave support after that branch is merged.  Unfortunately, I didn't have time to work on it recently.

PR #69, opened 2015-11-09
Add RAW argument to helm-bibtex-get-entry1

mwlodarczak:
The optional argument prevents calling helm-bibtex-prepare-entries. This
allows users who store paths to PDF files in a "File" field redefine
helm-bibtex-find-pdf without redefining helm-bibtex-get-entry1 itself.
The latter is neceessary to avoid infinite recursion:
helm-bibtex-get-entry1 calls helm-bibtex-prepare-entries, which in turn
calls helm-bibtex-find-pdf.

### DIFF
--- a/helm-bibtex.el
+++ b/helm-bibtex.el
@@ -321,7 +321,7 @@ appended to the requested entry."
                           :test (lambda (x y) (string= (s-downcase x) (s-downcase y)))
                           :key 'car :from-end t)))
 
-(defun helm-bibtex-get-entry1 (entry-key)
+(defun helm-bibtex-get-entry1 (entry-key &optional raw)
   (with-temp-buffer
     (mapc #'insert-file-contents
           (-flatten (list helm-bibtex-bibliography)))
@@ -329,8 +329,10 @@ appended to the requested entry."
     (re-search-forward (concat "^@\\(" parsebib--bibtex-identifier
                                "\\)[[:space:]]*[\(\{][[:space:]]*"
                                (regexp-quote entry-key) "[[:space:]]*,"))
-    (let ((entry-type (match-string 1)))
-      (reverse (helm-bibtex-prepare-entry (parsebib-read-entry entry-type))))))
+    (let* ((entry-type (match-string 1))
+          (entry (parsebib-read-entry entry-type)))
+      (reverse (or (and raw entry)
+                   (helm-bibtex-prepare-entry entry))))))
 
 (defun helm-bibtex-prepare-entries (entries)
   "Do some preprocessing of the entries."


tmalsburg:
Hi, I'm not convinced by that approach because, if I understand correctly, it means that every record is read two times which would lead to a considerable slowdown.  An alternative approach would be to change `helm-bibtex-find-pdf` such that it accepts a key or an entry.  There would have to be a customization variable `helm-bibtex-pdf-field` (or something like that) and if that variable is non-nil, `helm-bibtex-find-pdf` would first check that field before falling back to checking the library path. `helm-bibtex-prepare-entry` would then pass the entry to `helm-bibtex-find-pdf` but other functions could continue to call it with a key.  What do you think?

mwlodarczak:
Yep, that sounds better. I think "File" is a pretty standard name for such a field (that's what JabRef calls it, for instance) so that could even be used as the default value. I will look into it.

mwlodarczak:
OK, I have looked into it and while I like the idea of checking whether `helm-bibtex-pdf-field` is non-nil, I would rather have `helm-bibtex-find-pdf` take just keys as input and *if necessary* re-read the entry using my modified version of `helm-bibtex-get-entry1`.

If we make `helm-bibtex-find-pdf` accept keys or entries, we would need to modify `helm-bibtex-open-pdf`, `helm-bibtex-format-citation-org-link-to-PDF` and `helm-bibtex-add-PDF-attachment`, which would need to do exactly as much checking and re-reading as `helm-bibtex-find-pdf` below but it would be spread over three functions.

So what I am suggesting is something like:
```lisp
(defun helm-bibtex-find-pdf (key-or-entry)
  (if helm-bibtex-pdf-field
      (let ((entry (helm-bibtex-get-entry1 key t)))
        (helm-bibtex-get-value helm-bibtex-pdf-field entry))
    (-first 'f-exists?
            (--map (f-join it (s-concat key-or-entry ".pdf"))
                   (-flatten (list helm-bibtex-library-path))))))
```

Of course, the best solution would be to have some look-up table (and preferably have a "=pdf=" item with the path added alredy in `helm-bibtex-prepare-entry`) but I am not sure how well this would work with helm.

tmalsburg:
> If we make `helm-bibtex-find-pdf` accept keys or entries, we would need to modify `helm-bibtex-open-pdf`, `helm-bibtex-format-citation-org-link-to-PDF` and `helm-bibtex-add-PDF-attachment`,

I don't see that.  The idea behind making `helm-bibtex-find-pdf` accept a key or entry, was precisely that this allows us to leave these functions untouched.

> which would need to do exactly as much checking and re-reading as `helm-bibtex-find-pdf` 

Yes, these functions trigger some rereading, but they are usually only applied to a single entries, so the cost is very low.  Your solution rereads all entries on start-up and that will lead to a considerable slowdown for bigger bibliographies.

You are right that it would be nice if we could avoid rereading altogether.  The problem is that we would have to store the paths for all PDFs in memory although most of them will never be needed.  Some users have bibliographies with > 35k entries and in these cases storing all paths consumes a lot of memory and would lead to a noticeable slowdown during start-up.  Helm-bibtex' strategy is to initially read only the information that is absolutely necessary for doing the matching and to reread the entries when additional information is needed.  This allows us to deal with these gigantic bibliographies.

mwlodarczak:
On 11/10/2015 04:01 PM, Titus von der Malsburg wrote:
>> If we make `helm-bibtex-find-pdf` accept keys or entries, we would
>>  need to modify `helm-bibtex-open-pdf`, 
>> `helm-bibtex-format-citation-org-link-to-PDF` and 
>> `helm-bibtex-add-PDF-attachment`,
> 
> I don't see that.  The idea behind making `helm-bibtex-find-pdf` 
> accept a key or entry, was precisely that this allows us to leave 
> these functions untouched.

Perhaps I have missed something then. I mean, it will be
`helm-bibtex-open-pdf` will need to pass the entry to
`helm-bibtex-find-pdf` so it will need to re-read it as
helm-marked-candidates only includes keys. So `helm-bibtex-open-pdf`
would need to look more or less like `helm-bibtex-open-url-or-doi`.


> Your solution rereads all entries on start-up and that will lead to a
> considerable slowdown for bigger bibliographies.

You're right, of course. I missed the fact that `helm-bbitex-find-pdf`
is called already when generating candidates.

tmalsburg:
Upon thinking a little more about this: you are actually right and we do need to modify `helm-bibtex-get-entry1` or alternatively `helm-bibtex-prepare-entry`. ... I'll think about it.

tmalsburg:
I added a new branch [file-field](https://github.com/tmalsburg/helm-bibtex/tree/file-field).  Could you please test it?  The solution was a bit more involved than I anticipated and I would not be surprised if there were bugs lurking somewhere.

In this solution, I assumed that the `File` field stores absolute paths.  Is that correct?  (I don't use JabRef.)

mwlodarczak:
On 11/11/2015 03:17 AM, Titus von der Malsburg wrote:
> I added a new branch
> [file-field](https://github.com/tmalsburg/helm-bibtex/tree/file-field).
> Could you please test it?  The solution was a bit more involved than
> I anticipated and I would not be surprised if there were bugs lurking
> somewhere.

It seems to work fine, many thanks!

> In this solution, I assumed that the `File` field stores absolute
> paths.  Is that correct?  (I don't use JabRef.)

Yes. Actually, JabRef stores it a funny format like this:
{:/absolute/path/to/pdf.pdf:PDF} but I don't think we need to worry
about this. I will be happy to remove those extra elements from my
BibTeX file.

Marcin

tmalsburg:
[This (third screenshot)](http://blog.mendeley.com/tipstricks/howto-use-mendeley-to-create-citations-using-latex-and-bibtex/) suggests that Mendely may also produce this format.  I wonder if we should support it.  No one is going to edit all file fiels everytime they export their Mendeley or JabRef database.

tmalsburg:
Let's move this discussion to #71.

PR #74, opened 2015-11-19
Added full-frame variable

jagrg:
Git is very complicated. I hope I got right this time.


tmalsburg:
Looks good, but in the previous version you also had documentation in README.org.  Would you mind adding that?  Thank you.

jagrg:
Something went wrong again. Can you delete this?

tmalsburg:
I can only close the pull request but not delete it.

PR #77, opened 2015-11-19
Added option to set full-frame to nil

jagrg:
Sorry for the clutter, Titus. I think I might have overwritten the upstream/master branch. Anyway, it should work now.


tmalsburg:
Merged, thank you!

PR #80, opened 2015-11-25
Fix for searching higher level notes

jagrg:
My notes use level 2 headings. If you follow the structure below, try opening notes for Author 1 and let me know if you see the problem.

```
* Topic 1
** Author 1 (1999) Title
   :PROPERTIES:
   :Custom_ID: Key1
   :END:
* COMMENT
** Author 2 (2000) Title
   :PROPERTIES:
   :Custom_ID: Key2
   :END:

```

### DIFF
--- a/helm-bibtex.el
+++ b/helm-bibtex.el
@@ -873,10 +873,12 @@ defined.  Surrounding curly braces are stripped."
     (if (re-search-forward (format helm-bibtex-notes-key-pattern key) nil t)
                                         ; Existing entry found:
         (when (eq major-mode 'org-mode)
-          (outline-previous-visible-heading 1)
           (org-narrow-to-subtree)
-          (org-show-subtree)
-          (org-cycle-hide-drawers nil))
+          (save-restriction
+            (outline-previous-visible-heading 1)
+            (org-narrow-to-subtree)
+            (org-show-subtree)
+            (org-cycle-hide-drawers nil)))
                                         ; Create a new entry:
       (let ((entry (helm-bibtex-get-entry key)))
         (goto-char (point-max))


tmalsburg:
I don't understand what this code is supposed to fix.  In my Emacs, the current code works fine with your example.

jagrg:
The current code narrows Topic 1 as opposed to narrowing Author 1.

tmalsburg:
Ah, I see.  But I don't understand your code.  You use `org-narrow-to-subtree` within `save-restriction` which in my understanding reverses the narrowing.

tmalsburg:
I found a simpler fix, see b886d3c764862842ab92fbe218608eda774f4dd0.

jagrg:
You are right. I should've removed the second `org-narrow-to-subtree`. Anyhow, your commit didn't work too well for me. I tried a few times and Author 1 is not narrowing properly.

tmalsburg:
It works exactly as desired in my setup.  If you want me to fix it, you have to provide a precise description of the problem.

jagrg:
Make sure your notes file is collapsed. Now using the example above, do:

1. `M-x helm-bibtex`
2. search `Author 1`
3. press `f8` (Edit notes)

I get:

`|...`

If the file is expanded, I get

```
** Author 1 (1999) Title
   :PROPERTIES:|...
```

but this is what I would expect to see:

```
|** Author 1 (1999) Title
   :PROPERTIES:...
```

tmalsburg:
Ah, I see. Some of the commands did not work as expected when the point was placed within a folded subtree.  Can you please try bfcd5064dcc7c0ac62c46985832b2a73082f96e0?

jagrg:
It fixed the problem, but it leaves the entire buffer expanded.

tmalsburg:
What's wrong with that?

jagrg:
I have huge LOGBOOK drawers of clocked items which I don't want to see
open. https://github.com/jagrg/helm-bibtex/commit/5e3338c33ead0d3c212045922dda18be17139bb4 does a good job in keeping things tidy. Give it a try.

tmalsburg:
I'm guessing that you are storing your notes in a file that also contains other things.  Is that correct?  So far, I was assuming that people would use a separate file for notes (and in 99% of the cases that is probably correct).  I see why the expansion of the whole tree is annoying, in your case.

I'm wondering whether we're making our lives harder than necessary.  Org may already have higher-level infrastructure for the things we're doing, `org-refile` comes to mind and `org-capture`.  These do clearly not offer complete solutions but there may be more that we can use.

jagrg:
I'm actually using a separate file for notes as well. New notes are created under a top level COMMENT heading placed at the bottom of the file which I then refile to a different location. The thing is I'm often reviewing, sorting, editing and moving notes around. I don't just use it to retrieve notes that I wrote in the past is what I'm trying to say. I treat it like any other org file with a relatively coherent structure.

I'm not sure what the convention is for this kind of behaviour, but I would imagine something that preserves the buffer's default startup visibility (`org-set-startup-visibility`). If you open the agenda for example and press RET on one of the items, it opens the corresponding subtree without disturbing the rest of the buffer.

PR #92, opened 2016-01-27
Make buffer name unique

jagrg:
This is useful when `helm-resume` is called with a prefix ARG, which is used to select a different source from a list of buffers.

### DIFF
--- a/helm-bibtex.el
+++ b/helm-bibtex.el
@@ -342,7 +342,7 @@ containing authors, editors, title, year, type, and key of the
 entry.  This is string is used for matching.  The second element
 is the entry (only the fields listed above) as an alist."
   ;; Open configured bibliographies in temporary buffer:
-  (with-temp-buffer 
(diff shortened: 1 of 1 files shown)


tmalsburg:
Nice, I didn't know that helm-resume can restore different sessions.

jagrg:
>>>>> On 27 Jan 2016 at 19:52, Titus von der Malsburg <notifications@github.com> wrote:

> Nice, I didn't know that helm-resume can restore different sessions.

Yeah, until recently, neither did I. A similar trick I learned recently
is moving to the next source in a helm buffer with `C-o`. Quite useful
sometimes.

> --- Reply to this email directly or view it on GitHub:
> https://github.com/tmalsburg/helm-bibtex/pull/92#issuecomment-175822631

-- 
Jonathan

tmalsburg:
You can also use the left and right arrow keys to switch between sources, that's even easier.

PR #94, opened 2016-02-15
Support for customising which fields to display in the *helm bibtex* buffer

kintism:
Introduced `helm-bibtex-display-format' variable whose customisation allows the user to display different fields in the *helm bibtex* buffer and control how they are formatted. 

For example, the following settings will display note availability, key, title and authors per entry:

```
(setq helm-bibtex-display-format
      '((fields-to-display . ("=has-note=" "=key=" "title" "author"))
        (fields-format     . "$0 $1 $2 $3")
        (fields-width      . (lambda (width) (list 2 25 (- width 65) 35)))))
```

### DIFF
--- a/helm-bibtex.el
+++ b/helm-bibtex.el
@@ -314,6 +314,33 @@ nil, the window will split below."
   :group 'helm-bibtex
   :type 'boolean)
 
+(defcustom helm-bibtex-display-format
+  '((fields-to-display . ("author" "title" "year"
+                          "=has-pdf=" "=has-note=" "=type="))
+    (fields-format     . "$0 $1 $2 $3$4 $5")
+    (fields-width      . (lambda (width) (list 36 (- width 53) 4 1 1 7))))
+  "Fields to display in the *helm bibtex* buffer and their format.
+
+An alist with the following KEYs:
+ - fields-to-display: a list of fields to display
+ - fields-format:     their format as accepted by `s-format'
+ - fields-width:      a function that takes as argument the
+                      width of the helm window and returns a
+                      list of widths per displayed field
+
+Some examples:
+ 1. Display key, title and note availability per entry:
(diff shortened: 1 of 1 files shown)


tmalsburg:
Thanks for the PR.  I like the idea and wanted to add something similar for a long time.  I think it would be a good to factor out the code for table formatting into a separate function.  This function could also be used in other helm sources.  Also, the defcustom could be simplified to something like the following:

``` elisp
(setq helm-bibtex-display-format
     '((fields-to-display . ("=has-note=" "=key=" "title" "author"))
       (fields-width      . (2 25 (- width 65) 35))))
```

The order of fields would then be determined by the order in `fields-to-display`.  What do you think?  I understand if you don't want to work on this, because the code that you have solves your immediate problem.

kintism:
Hi Titus, thank you for your reply. Your suggestion of a general formatter is
very interesting. I think it can be easily accomplished for helm-bibtex, but a
general solution seems more difficult. Although I am not a helm expert, I think
we will need access to the individual "fields" of a source's candidates; I don't
know if such information is available for a given source. In any case, I will
look into it and let you know.

Regarding the simplification of the defcustom, I believe the `fields-format` would
be valuable to a user. As an example, consider the following case:

```emacs-lisp
(setq helm-bibtex-display-format '((fields-to-display . ("=has-pdf=" "=has-note=" "author" "year" "title"))
                                   (fields-format     . "$0$1 $2 ($3) $4")
                                   (fields-width      . (lambda (width) (list 1 1 35 4 (- width 50))))))
```

tmalsburg:
The function for formatting would probably take the candidates as a lists of alists and other helm sources would have to use that format, too. It may be worth looking at Mu4e's code for formatting the inbox.  It is configurable, too, and we may be able to recycle a lot of the code and ideas.

I agree about `fields-format`, it's useful to have it.

PR #98, opened 2016-03-08
Added minor mode for editing notes

jagrg:
I've added a few key bindings that I think makes the workflow smoother. The idea comes from org-capture. Basically, `C-c C-c` to exit the buffer, `C-c C-h` to return to the helm bibtex buffer and `C-c C-w` to refile. The target level can be changed by the user: `(setq org-refile-targets '((helm-bibtex-notes-path :maxlevel . 1)))`.

I've also added a new line when creating new entries in the bib file (see `helm-bibtex-fallback-candidates`).

### DIFF
--- a/helm-bibtex.el
+++ b/helm-bibtex.el
@@ -858,6 +858,44 @@ defined.  Surrounding curly braces are stripped."
       (-each it 'mml-attach-file)
     (message "No PDF(s) found.")))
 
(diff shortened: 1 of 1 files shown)


tmalsburg:
Thank you.  This looks like a very useful addition.  I'll test it a bit.

PR #102, opened 2016-03-19
Make generic library and add ivy support

justbur:
New file bibtex-completion.el holds a generic version of
helm-bibtex.el. Any helm specific functions are used by setting
variables, all of which are set automatically by the
bibtex-completion-frontend option.

New file ivy-bibtex.el holds ivy-bibtex which is the ivy translation of
helm-bibtex.

No changes are made in helm-bibtex.el

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -1,0 +1,1017 @@
+;;; bibtex-completion.el --- A BibTeX bibliography manager based on Helm
+
+;; Copyright 2014 Titus von der Malsburg <malsburg@posteo.de>
+
+;; Author: Titus von der Malsburg <malsburg@posteo.de>
+;; Maintainer: Titus von der Malsburg <malsburg@posteo.de>
+;; Version: 1.0.0
+;; Package-Requires: ((parsebib "1.0") (s "1.9.0") (dash "2.6.0") (f "0.16.2") (cl-lib "0.5"))
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; A BibTeX bibliography manager based on Helm.
+;;
+;; News:
+;; - 02/25/2016: Support for pre- and postnotes for pandoc-citeproc
+;;   citations.
+;; - 11/23/2015: Added support for keeping all notes in one
+;;   org-file.  See customization variable `bibtex-completion-notes-path'.
+;; - 11/10/2015: Added support for PDFs specified in a BibTeX
+;;   field.  See customization variable `bibtex-completion-pdf-field'.
+;; - 11/09/2015: Improved insertion of LaTeX cite commands.
+;;
+;; See NEWS.org for old news.
+;;
+;; Key features:
+;; - Quick access to your bibliography from within Emacs
+;; - Tightly integrated workflows
+;; - Provides instant search results as you type
+;; - Powerful search expressions
+;; - Open the PDFs, URLs, or DOIs associated with an entry
+;; - Insert LaTeX cite commands, Ebib links, or Pandoc citations,
+;;   BibTeX entries, or plain text references at point, attach PDFs to
+;;   emails
+;; - Attach notes to publications
+;; - Quick access to online bibliographic databases such as Pubmed,
+;;   arXiv, Google Scholar, Library of Congress, etc.
+;;
+;; See the github page for details:
+;;
+;;    https://github.com/tmalsburg/helm-bibtex
+
+;;; Install:
+
+;; Put this file in a directory included in your load path or install
+;; bibtex-completion from MELPA (preferred).  Then add the following in your
+;; Emacs startup file:
+;;
+;;     (require 'bibtex-completion)
+;;
+;; Alternatively, you can use autoload:
+;;
+;;     (autoload 'bibtex-completion "bibtex-completion" "" t)
+;;
+;; Requirements are parsebib, helm, s, dash, and f.  The easiest way
+;; to install these packages is through MELPA.  Make sure helm is
+;; properly configured (see
+;; https://github.com/emacs-helm/helm#install-from-emacs-packaging-system).
+;;
+;; Let bibtex-completion know where it can find your bibliography by setting
+;; the variable `bibtex-completion-bibliography'.  See the manual for more details:
+;;
+;;   https://github.com/tmalsburg/helm-bibtex#minimal-configuration
+
+;;; Usage:
+
+;; You can search entries using the command `bibtex-completion'.  Select an
+;; entry and press TAB to access all available actions.  At the end of
+;; the list of matches you find some dummy entries that can be used
+;; for searching in online databases.  Apart from that, familiarize
+;; yourself with Helm.  It's more powerful that you might think.
+
+;;; Code:
+
+(require 'browse-url)
+(require 'parsebib)
+(require 'cl-lib)
+(require 'dash)
+(require 's)
+(require 'f)
+
+(defgroup bibtex-completion nil
+  "Helm plugin for searching entries in a BibTeX bibliography."
+  :group 'completion)
+
+(defcustom bibtex-completion-bibliography nil
+  "The BibTeX file or list of BibTeX files."
+  :group 'bibtex-completion
+  :type '(choice file (repeat file)))
+
+(defcustom bibtex-completion-library-path nil
+  "A directory or list of directories in which PDFs are stored.
+Bibtex-completion assumes that the names of these PDFs are
+composed of the BibTeX-key plus a \".pdf\" suffix."
+  :group 'bibtex-completion
+  :type '(choice directory (repeat directory)))
+
+(defcustom bibtex-completion-pdf-open-function 'find-file
+  "The function used for opening PDF files.  This can be an
+arbitrary function that takes one argument: the path to the PDF
+file.  The default is `find-file' which opens the PDF in
+Emacs (either with docview or, if installed, the much superior
+pdf-tools.  When set to `helm-open-file-with-default-tool', the
+systems default viewer for PDFs is used."
+  :group 'bibtex-completion
+  :type 'function)
+
+(defcustom bibtex-completion-pdf-symbol "⌘"
+  "Symbol used to indicate that a PDF file is available for a
+publication.  This should be a single character."
+  :group 'bibtex-completion
+  :type 'string)
+
+(defcustom bibtex-completion-format-citation-functions
+  '((org-mode      . bibtex-completion-format-citation-ebib)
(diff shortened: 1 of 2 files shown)


tmalsburg:
Nice work!  I will play with it a little before merging.

justbur:
No problem. I didn't have much time to test yet, so this is still a WIP. Doesn't touch the existing package though, so I thought there was no harm

tmalsburg:
Ok, I had a quick look through the code.  Looks very clean and more straight-forward than I thought.  Thank you very much!  Some comments are below.

It would be nice if the back-end was completely agnostic to the front-end, i.e., if `bibtex-completion-frontend` et al. weren’t needed.  I think this could be achieved by adding a thin layer on top of the actions in the front-end.  In the case of helm-bibtex, an example would be:


``` elisp
(defun helm-bibtex-insert-citation (_)
  (let ((keys (helm-marked-candidates :with-wildcard t))
    (with-helm-current-buffer
      (bibtex-completion-insert-citation keys)))))
```

For ivy-bibtex, it would look something like this:

``` elisp
(defun ivy-bibtex-insert-citation (candidate)
  (bibtex-completion-insert-citation (list candidate)))
```

and we could change `bibtex-completion-insert-citation` such that even this facade isn’t needed (just wrap an argument in a list if it isn’t a list already).

Question: Am I correct in assuming that ivy can execute actions only for one candidate at once?  If yes, that would be an argument for using helm-bibtex, where I select multiple candidates all the time.  For example, I enter my name, select all candidates, and chose “Insert reference” and have a full bibliography of my work.  Easy.

Another issue: you sneak the formatting into bibtex-completion-candidates, presumably to have the formatted entries in the cache.  Caching the formatted entries is actually a good idea but I wonder whether we can find a cleaner way to do this.  I think helm’s approach to separate the loading of the candidates and the formatting is pretty smart for a number of reasons.  One is that formatting can take a long time for large bibliographies (e.g., CryptoBib, https://cryptobib.di.ens.fr/).  Helm can be configured to format only the candidates that are displayed (`filtered-candidate-transformer`) and this is of course much faster.  A second reason is that formatting can truncate the title or the author list and if ivy matches on the formatted candidates you may not find an entry because the crucial part was cut off.  Unfortunately, I don't know enough about ivy to tell how this could be approach.

Thanks again.  This is really cool stuff.

justbur:
> It would be nice if the back-end was completely agnostic to the front-end, i.e., if bibtex-completion-frontend et al. weren’t needed. I think this could be achieved by adding a thin layer on top of the actions in the front-end. In the case of helm-bibtex, an example would be:

That's fine with me, but I think there are a couple of points where you need to know that it's helm. By the way, I've seen a couple of libraries have a variable to specify a completion tool, so that's why I thought of this. projectile is one I think.

There's the `with-helm-current-buffer` thing that I don't know how to deal with nicely. In ivy, actions are executed in the current buffer already so this is unnecessary. I'm not very familiar with how helm works behind the scenes. 

> Question: Am I correct in assuming that ivy can execute actions only for one candidate at once? If yes, that would be an argument for using helm-bibtex, where I select multiple candidates all the time. For example, I enter my name, select all candidates, and chose “Insert reference” and have a full bibliography of my work. Easy.

Yep, that's true, and I like the interface you have here. I'm just not a helm person, so I think it's nice to have your work available for the rest of us. What you can do with ivy which is almost the same is execute an arbitrary action on a candidate without exiting, which is almost the same as selecting then executing an action on multiple candidates. For most applications, I don't miss this helm feature, but that's just me. 

> Another issue: you sneak the formatting into bibtex-completion-candidates, presumably to have the formatted entries in the cache. Caching the formatted entries is actually a good idea but I wonder whether we can find a cleaner way to do this. I think helm’s approach to separate the loading of the candidates and the formatting is pretty smart for a number of reasons. One is that formatting can take a long time for large bibliographies (e.g., CryptoBib, https://cryptobib.di.ens.fr/). Helm can be configured to format only the candidates that are displayed (filtered-candidate-transformer) and this is of course much faster. A second reason is that formatting can truncate the title or the author list and if ivy matches on the formatted candidates you may not find an entry because the crucial part was cut off. Unfortunately, I don't know enough about ivy to tell how this could be approach.

Let me play with this. Ivy has a transformer option, but I think it only acts on individual candidates. I tried to translate this one, but it didn't work easily.  I didn't notice any slowdown on my bib file, but it's not that large (~200 entries).

tmalsburg:
> There's the `with-helm-current-buffer` thing that I don't know how to deal with nicely.

I think my wrapper (see last comment) deals with that, but I haven't tried it yet (working on it).

Re: helm vs ivy, I just asked because I was curious, not judging.  If other users prefer ivy, that's perfectly fine with me.

Re formatter: My bib has ~1100 entries and this is were you start noticing when stuff is slow.  CryptoBib has > 35,000 entries.

justbur:
> I think my wrapper (see last comment) deals with that, but I haven't tried it yet (working on it).

That may be cleaner yes. This should probably just be a library.

> Re: helm vs ivy, I just asked because I was curious, not judging. If other users prefer ivy, that's perfectly fine with me.

No worries

> Re formatter: My bib has ~1100 entries and this is were you start noticing when stuff is slow. CryptoBib has > 35,000 entries.

There's another issue with the ivy formatter and that's if the formatting is done once the frame size might change and screw up the formatting. If during parsing you stored the width of the columns you could use the transformer option in ivy that only knows about the current candidate. I can ask the ivy developer what he would do in this instance.

tmalsburg:
The backend is now complete agnostic to the frontend (if I haven't overlooked something): d2f747a079936cff75f0d3e10dcde4b178461af7

I also added a first version of helm-bibtex based on the backend: https://github.com/tmalsburg/helm-bibtex/blob/master/helm-bibtex-v2.el

So far it seems to work well.  The facades in helm-bibtex can actually be built with a macro, so that part will be simplified.  Nice!

tmalsburg:
I think we can also share the code for the formatter.  AFAICS the only difference is how we calculate the window-width and that can also be done in a wrapper in helm-bibtex and ivy-bibtex.

justbur:
Nice! I posted a question to the maintainer of ivy about the formatter issue. Let's see what he says about it. 

I think I'd put two recipes up on melpa for this, one each for helm and ivy, both of which use the backend. Or you could have three and add a dependency.

tmalsburg:
The formatter is now generic: a4a67103b7cd91d04ef17583a8c0979486f1b90f

tmalsburg:
If the helm and ivy recipes both include the backend, do we get in trouble when someone installs both helm-bibtex and ivy-bibtex?  I would think not, but I'm not sure.

justbur:
Looking good :)

[Here's another option](https://github.com/abo-abo/swiper/issues/447#issuecomment-198862757), but it doesn't quite work with your formatting function. What this does is to pass the unformatted cars of the candidates to the formatting function. Just fyi

> If the helm and ivy recipes both include the backend, do we get in trouble when someone installs both helm-bibtex and ivy-bibtex? I would think not, but I'm not sure.

There might be an issue if someone gets two versions of bibtex-completion.el installed. Say they installed helm-bibtex a while ago and update ivy-bibtex without helm-bibtex. It's possible they end up loading an out of date bibtex.completion file. With three separate packages you can use required versions to force the bibtex-completion to be upgraded. However, I think magit had problems with this as it was possible to get into a state where melpa did not have the required version available for a few hours. So I guess I'm saying there might be drawbacks to both approaches. 

> Their configurations should ideally continue to work although the configuration variables are renamed. Not sure how to do this, though.

I think the "right" way to do this is to translate the old names to the new names, but mark the old names as obsolete so they get a warning. You could probably do most of it with `defvaralias` and `make-obsolete-variable`. [Here's an example of a way I handled a name change](https://github.com/justbur/emacs-which-key/blob/master/which-key.el#L274-L284).

tmalsburg:
Thanks for your input on defvaralias.  I tried the code below but it doesn't work (no error message).  I probably misunderstand something about macros or symbols or their interaction.  Hm …

``` elisp
(cl-loop
 for var in '("bibliography" "library-path" "pdf-open-function"
              "pdf-symbol" "format-citation-functions" "notes-path"
              "notes-template-multiple-files"
              "notes-template-one-file" "notes-key-pattern"
              "notes-extension" "notes-symbol" "fallback-options"
              "browser-function" "additional-search-fields"
              "no-export-fields" "cite-commands"
              "cite-default-command"
              "cite-prompt-for-optional-arguments"
              "cite-default-as-initial-input" "pdf-field")
 for oldvar = (make-symbol (concat "helm-bibtex-" var))
 for newvar = (make-symbol (concat "bibtex-completion-" var))
 do
 (defvar oldvar)
 (defvaralias newvar oldvar)
 (make-obsolete-variable oldvar newvar "2016-03-20"))

```

tmalsburg:
> Here's another option, but it doesn't quite work with your formatting function. What this does is to pass the unformatted cars of the candidates to the formatting function. Just fyi

So we'd have to create function `ivy-bibtex-candidates-formatter` that preprocesses the candidate list a little before passing it to `bibtex-completion-candidates-formatter`.  (ivy passes a list of pairs but `bibtex-completion-candidates-formatter` expects a list of just the alists.  Anything else?

tmalsburg:
Re MELPA: It is possible that org-ref's @jkitchin wants to use the backend but not the frontend.  So perhaps, three MELPA receipts are better.

justbur:
Try this
```elisp
(cl-loop
 for var in '("bibliography" "library-path" "pdf-open-function"
              "pdf-symbol" "format-citation-functions" "notes-path"
              "notes-template-multiple-files"
              "notes-template-one-file" "notes-key-pattern"
              "notes-extension" "notes-symbol" "fallback-options"
              "browser-function" "additional-search-fields"
              "no-export-fields" "cite-commands"
              "cite-default-command"
              "cite-prompt-for-optional-arguments"
              "cite-default-as-initial-input" "pdf-field")
 for oldvar = (intern (concat "helm-bibtex-" var))
 for newvar = (intern (concat "bibtex-completion-" var))
 do
 (defvaralias oldvar newvar)
 (make-obsolete-variable oldvar newvar "2016-03-20"))
```
I'm not too familiar with `cl-loop` though

justbur:
> So we'd have to create function ivy-bibtex-candidates-formatter that preprocesses the candidate list a little before passing it to bibtex-completion-candidates-formatter. (ivy passes a list of pairs but bibtex-completion-candidates-formatter expects a list of just the alists. Anything else? 

For now, I think the current method is fine. I appreciate that you are trying to get it just right (specifically right aligning the columns), but another more limited approach on the ivy side would be to come up with reasonable fixed column sizes and let the remaining results be truncated. I personally wouldn't mind if some of the information was truncated when the frame is small as long as I know that I can enlarge it if I need to. 

A slightly different way to do it would be to guess reasonable "column widths" based on the current frame size then truncate each column to that width. This would allow the transformer to act independently on each candidate.

Another alternative would be to come up with a reasonable modification to the way ivy works that would make this easier. I think as long as its not too complicated we could add it.

> Re MELPA: It is possible that org-ref's @jkitchin wants to use the backend but not the frontend. So perhaps, three MELPA receipts are better.

sounds good to me

jkitchin:
Org-ref has taken a similar path to specifying back ends including reftex, helm-bibtex, an internal helm-cite and an ivy backend. We only use them for selection and cite link insertion.  The point is we can adapt as things change ;)

tmalsburg:
@jkitchin, you may have to make a few changes when we switch to the new code (currently experimental).  But it should be as simple as renaming some functions calls, no big deal.  I'll keep you posted.

tmalsburg:
> A slightly different way to do it would be to guess reasonable "column widths" based on the current frame size then truncate each column to that width. This would allow the transformer to act independently on each candidate.

The problem of formatting tables arises in many contexts and it would be fantastic to have a general library for that.  I hope that I will find time to implement something like this at some point.

PR #103, opened 2016-03-21
ivy-bibtex: Add heuristic transformer

justbur:
Here's a different approach to the formatting. The idea is to use a percentage of the available width for authors, fixed sizes for year and type, and the remaining width for the title. 

The pre-process thing is a hack right now. If you like this I can make this string the default so the transformer works agnostically.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -295,15 +295,31 @@ before being saved."
                (entries (bibtex-completion-prepare-entries entries))
                (entries (nreverse entries))
                (entries
-                (--map (cons (bibtex-completion-clean-string
-                                  (s-join " " (-map #'cdr it))) it)
-                           entries)))
+                (mapcar #'bibtex-completion-add-candidate-string
+                        entries)))
           (setq bibtex-completion-cached-candidates
                 (if (functionp formatter)
                     (funcall formatter entries)
                   entries)))
         (setq bibtex-completion-bibliography-hash bibliography-hash))
       bibtex-completion-cached-candidates)))
+
+(defun bibtex-completion-add-candidate-string (entry)
+  "Add string describing candidate to head of ENTRY. This string
+is specially formatted for `ivy-bibtex-display-transformer'."
+  (cons
+   (mapconcat
+    (lambda (field)
+      (let ((res (bibtex-completion-clean-string
+                  (bibtex-completion-get-value field entry " "))))
+        (if (member field '("author" "editor"))
+            (bibtex-completion-shorten-authors res)
+          res)))
+    (if (assoc-string "author" entry 'case-fold)
+        '("author" "title" "year" "=has-pdf=" "=has-note=" "=type=")
+      '("editor" "title" "year" "=has-pdf=" "=has-note=" "=type="))
+    "\t")
+   entry))
 
 (defun bibtex-completion-resolve-crossrefs (entries)
   "Expand all entries with fields from cross-references entries."
--- a/ivy-bibtex.el
+++ b/ivy-bibtex.el
@@ -79,10 +79,62 @@
 
 (require 'ivy)
 (require 'bibtex-completion)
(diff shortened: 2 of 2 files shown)


justbur:
updated with less code duplication

tmalsburg:
I like the idea to make the author field dynamic-width.  Some comments:

I think `bibtex-completion-clean-string` isn't needed for the key in `ivy-bibtex--pre-process-candidates`.

`ivy-bibtex--pre-process-candidates` ignores `bibtex-completion-additional-search-fields`.  So the user cannot extend the fields that they want to use for searching.  (Additional search fields, such as abstracts, are not displayed in helm-bibtex.)

You can use `bibtex-completion-shorten-authors` to the reduce the author list to last names.

Is there a way to include symbols for PDFs and notes (`=has-pdf=` and `=has-notes=`)?  Perhaps the preprocessor can simply include these strings (`=has-pdf=`, `=has-notes=`).  And the transformer can replace them with the symbols.  This way people can even search for entries that have PDFs or notes.

More generally, the preprocessor returns conses of strings used for matching and the key.  Could it instead return conses of the string for searching and the alist representing the entry?  The transfomer could then perhaps use the alist to construct the display instead of having to parse the string for matching.  (This is the approach taken by helm-bibtex.)

justbur:
> Could it instead return conses of the string for searching and the alist representing the entry? The transfomer could then perhaps use the alist to construct the display instead of having to parse the string for matching. (This is the approach taken by helm-bibtex.) One benefit would of course be that we could use the same transformer for helm-bibtex and ivy-bibtex.

The issue here is that at least in the way it currently works the basic ivy transformer only "sees" the candidate string. I know embedding the tabs in the string to use for formatting is not great, but that's why I did it this way. I can ask about getting this changed

tmalsburg:
I see. Not sure whether it's worth making changes to ivy just to allow us to simplify our code a little.

I suppose, one option is that we split the candidate string and construct an alist that we then pass to the generic transformer. Still a hack but less code duplication.

justbur:
If I remember right you were not using the string generated at the parsing stage at all. Why not just make that one one that ivy can use along these lines?

tmalsburg:
The string generated at the parsing stage is used by Helm for matching.  But yes, as long as it contains all the necessary information (including `bibtex-completion-additional-search-fields`) it doesn't really matter what format it has, and using the tab-separated format would be fine.

justbur:
Crap this screwed up the actions, because now it skips the step where the cdr is made into just the key.

tmalsburg:
Sorry for the silence, I was on vacation over Easter.  What's the status of ivy-bibtex?  I have been using the new version of helm-bibtex for a while now and it works really well, so I would feel comfortable releasing it to MELPA.

justbur:
No problem. To get this transformer to work, I would have to change the way the helm side works, too, so I stopped working on it. Besides the fact that all of the candidates for ivy need to be processed up front, I think it works fine without this pr.

tmalsburg:
So shall we put up a MELPA recipe for ivy-bibtex?

justbur:
> So shall we put up a MELPA recipe for ivy-bibtex?

Sure, seems to be ready to me. If there are performance issues, we can fix them later.

tmalsburg:
Ok, I'm rewriting the documentation in README.org.  One question: how can I select actions in ivy-bibtex?

justbur:
Feel free to change the default action. 

[Here is a link to the docs](http://oremacs.com/swiper/#key-bindings-for-single-selection-action-then-exit-minibuffer), which probably explain it better than I can.

tmalsburg:
Steve merged the [ivy-bibtex](https://github.com/melpa/melpa/pull/3773) recipe into MELPA (but it doesn't show up in the index yet).

justbur:
Nice, I'll keep an eye out for it. Thanks

PR #113, opened 2016-06-27
Determine BibTeX database locally

jagrg:


tmalsburg:
Thank you for the PR.  I'm currently tied up with some projects and will hopefully be able to look at it on the weekend.

ericdanan:
I find this very useful and would have two suggestions (both inspired by the function `org-ref-find-bibliography` in the `org-ref` package):

1. If we call `ivy-bibtex-with-local-bibliography` (or the `helm-` analogue) from inside a `bib` file, use this file as local bibliography (this gives a quick way to navigate a bibtex file).
2. `reftex` has a function `reftex-locate bibliography-files` that finds the relevant `bib` files in a more robust way I think: it can catch multiple `\addbibresource` that are allowed with `biblatex`, find `\addbibresource` or `\bibliography` in all project files and not only the master file, and locate the files when environment variables like `BIBINPUTS` are used.

Here is a simple implementation of these two suggestions that works for me:

```
(defun bibtex-completion--get-local-databases ()
  "Return a list of .bib files associated with the current file."
  (cond
   ((string= (or (f-ext (or (buffer-file-name) "")) "") "bib")
    (list (buffer-file-name)))
   ((bound-and-true-p reftex-mode)
    (reftex-locate-bibliography-files
     (if (fboundp 'TeX-master-directory)
	 (TeX-master-directory)
       (or (file-name-directory (or (buffer-file-name) ""))
	   bibtex-completion-bibliography))))
   (t
    bibtex-completion-bibliography)))
```

Perhaps you may want to add another case before the last one in the `cond` to cater for people using `auctex` or `org-mode` without `reftex`. In this case I imagine you can't do much more than searching "by hand" as you do. But I think keeping the `reftex` case beforehands would still be better for `reftex` users.

Also I have the impression that once `bibtex-completion--get-local-databases` is defined, you can simply do:

```
;;;###autoload
(defun ivy-bibtex-with-local-bibliography (&optional arg)
  "Search BibTeX entries with local bibliography.

With a prefix ARG the cache is invalidated and the bibliography reread."
  (interactive "P")
  (let ((bibtex-completion-bibliography (bibtex-completion--get-local-databases)))
    (ivy-bibtex arg)))
```

and similarly for `helm`.

Anyway, thanks a lot and I'm happy to test more things if you need (not sure I can / should add commits to this PR though).

ericdanan:
Sorry, I didn't do it right for org-mode. Indeed the usual way to use `reftex` in org-mode is as follows:

```
(defun org-mode-reftex-setup ()
  "Setup reftex for org-mode."
  (load-library "reftex")
  (and (buffer-file-name)
       (file-exists-p (buffer-file-name))
       (reftex-parse-all)))
 
(add-hook 'org-mode-hook 'org-mode-reftex-setup)
```

This loads the `reftex` library but does not turn on `reftex-mode`. So my proposal above will not pick up the local databases. But the one below works:

```
(defun bibtex-completion--get-local-databases ()
  "Return a list of .bib files associated with the current file."
  (or (and (buffer-file-name)
           (string= (or (f-ext (buffer-file-name)) "") "bib")
           (list (buffer-file-name)))
      (and (buffer-file-name)
           (fboundp 'reftex-locate-bibliography-files)
           (reftex-locate-bibliography-files
            (if (fboundp 'TeX-master-directory)
                (TeX-master-directory)
              (file-name-directory (buffer-file-name)))))
      bibtex-completion-bibliography))
```

jagrg:
@ericdanan Thanks. I'm not too familiar with point 2, but it looks fine. You've added an extra hook. Is this something the user would need to add?

@tmalsburg Can you send us an update?

ericdanan:
Adding `org-mode-reftex-setup` to `org-mode-hook` is necessary if the user wants reftex to pick up the local databases in org-mode files. But this is not specifically for bibtex-completion, it seems to be the most usual way to use reftex in org-mode files and is recommended in this [org-mode FAQ](http://orgmode.org/worg/org-faq.html#using-reftex-in-org-mode).

Now if somebody wants this to work in org-mode files and did not setup reftex for org-mode, I guess we should either point her to the FAQ to setup reftex, or add another clause before the last line to search "by hand" as you proposed (in which case I think we can forget about the master file issue and simply search the current file, but on the other hand we may want to search for all occurences of `\addbibresource`).

tmalsburg:
Hi both, thanks for your work on this.  This looks useful indeed.  There is one thing that worries me a bit, though: using the with-local variants of helm-bibtex and ivy-bibtex invalidates the cache for standard helm- and ivy-bibtex, which could be annoying for people working with large bibliographies.  Is there an easy way to avoid this?

I like @ericdanan's suggestions but this would add reftex as another dependency.  Is reftex part of standard Emacs? In this case, it wouldn't matter much.

ericdanan:
According to the [reftex manual](https://www.gnu.org/software/emacs/manual/html_node/reftex/Installation.html), reftex is part of emacs since version 20.2 (I didn't know that, thought it was part of the auctex package). So I think it is even better to replace

    (fboundp 'reftex-locate-bibliography-files)

by

    (require 'reftex-parse nil t)

in the function, so that reftex will be used systematically. And the `org-mode-hook` setup then becomes unnecessary, the function will actually work in any buffer in which we insert the standard latex `\bibliography` or `\addbibresource`. The library loads instantly so no issure here.

To avoid the cache invalidation issue, we can easily separate a global and a local cache. I have tried it and checked with a huge bib file (crypto.bib) that it does not impact loading time. Since this implies a couple of changes in bibtex-completion.el as well, I think it will be more convenient if I make a separate PR so you can easily test things.

tmalsburg:
I'm closing this. Thanks, @jagrg, for getting this going!

PR #125, opened 2016-08-30
Widen notes file

jagrg:
1. M-x helm-bibtex
2. edit notes (org-narrow-to-subtree if not already narrowed)
3. C-u M-x helm-bibtex
4. (notes symbols are not shown)


tmalsburg:
Good catch, thank you!

PR #127, opened 2016-09-02
make ivy-bibtex default action easily customizable

ericdanan:
- Added a custom variable ivy-bibtex-default-action, so that the ivy-bibtex default action can be changed without redefining the whole ivy-bibtex-function.
- Updated the readme acordingly.
- Also fixed a typo (I think) in helm-bibtex.el.


tmalsburg:
Thank you very much, Eric!

PR #128, opened 2016-09-02
add fallback options to ivy-bibtex

ericdanan:
This is done by introducing a new action offering to select a fallback
options (this is the easiest way I could think of, since ivy-read has no
builtin mechanism to simultaneously use different sources with different
matchers and actions).

To do this efficiently I made the search expression a variable of the
functions in bibtex-completion-fallback-options (helm-pattern or
ivy-text, depending on the backend). I updated the readme accordingly.

I tested with ivy and helm and everything seems to work.

### DIFF
--- a/README.org
+++ b/README.org
@@ -9,6 +9,7 @@ Helm-bibtex and ivy-bibtex allow you to search and manage your BibTeX bibliograp
 Ivy-bibtex is experimental at this time, lacks some features, and may be noticeably slower than helm-bibtex when used with larger bibliographies.
 
 * News
+- [2016-09-20 Tue] :: Added fallback options to ivy frontend.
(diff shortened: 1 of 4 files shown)


tmalsburg:
Looks good, thanks you! Could you please rename `string` to something more informative like `search-expression`?  I suppose it would also be worth announcing this feature in the [news section in README.org](https://github.com/tmalsburg/helm-bibtex#news) and the [news section in ivy-bibtex.el](https://github.com/tmalsburg/helm-bibtex/blob/master/ivy-bibtex.el#L27) (this is used by MELPA).

ericdanan:
Done.

tmalsburg:
Awesome. Thanks for the contribution.  Very much appreciated!

PR #130, opened 2016-09-09
"smart" formatter for latex citations

ericdanan:
Added a new function `bibtex-completion-format-citation-cite-dwim`. It uses `reftex` (if turned on) to determine if point is inside (or right after) a citation macro. If so it only adds keys to the existing citation, otherwise it calls `bibtex-completion-format-citation-cite` to insert a new citation.

I think it is especially useful for `ivy-bibtex`, because it makes it possible to quickly cite a bunch of references at once, using `ivy-call`. Otherwise one has to call `ivy-bibtex` and insert the first citation using`ivy-done`, then manually go back one char and insert a comma, then call `ivy-bibtex` again to insert the second key, manually insert a comma, call `ivy-bibtex` once agaiin, and so on.

PS sorry if I am sending you too many PRs...

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -609,6 +609,43 @@ for arguments if the commands can take any."
             (format "\\%s{%s}" cite-command (s-join ", " keys))
           (format "\\%s[%s][%s]{%s}" cite-command prenote postnote (s-join ", " keys)))))))
 
+(defun bibtex-completion-format-citation-cite-dwim (keys)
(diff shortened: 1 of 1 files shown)


tmalsburg:
This looks useful but I think it would make sense to use this function by default instead of bibtex-completion-format-citation-cite.

I only had a quick look at the code.  Does it work with other cite macros than `cite`, e.g., `parencite`, `citeNP`, etc.?

tmalsburg:
Sorry for the slow response, btw, and thanks for your contributions! :-)

ericdanan:
Done (I hope this is the way you wanted it, rather than keeping two formatter functions I incorporated the code from the -dwim formatter into the default one).

Yes it works with any latex macro whose name contains "cite".

tmalsburg:
Fantastic.  Looking forward to using this feature.

PR #137, opened 2016-09-20
Local bibliography (contd from #113)

ericdanan:
/cc @jagrg who wrote PR #113 

So as I commented there, I think it is more convenient at that point to send you a new PR.

There are two commits.

The first one simply adds the function `bibtex-completion--get-local-databases` from #113 (I took the liberty to rename it to `bibtex-completion-find-local-bibliography` which I think is more in line with the way you name functions in the package...), with the small change from my last comment, as well as the `-with-local-bibliography` variants of `helm-bibtex` and `ivy-bibtex`.

The second commit implements a separate local cache so that the `-with-local` variants don't invalidate the cache for the standard ones. The idea is simply to replace the variables `bibtex-completion-bibliography-hash` and `bibtex-completion-cached-candidates` with an alist (named `bibtex-completion-cache`) of the form

    (('global HASH CANDIDATES) ('local HASH CANDIDATES))

and to add a variable `bibtex-completion-bibliography-type`  which is `'global` by default but is let-bound to `'local` by the `-wich-local` variants. Then we only need to change a couple of lines so that `bibtex-completion-candidates` uses this latter variable to read and write at the right place in the cache.

I measured loading time with crypto.bib and the following code

```
(defmacro measure-time (&rest body)
  "Measure the time it takes to evaluate BODY."
  `(let ((time (current-time)))
     ,@body
     (message "%.06f" (float-time (time-since time)))))
(require 'bibtex-completion)
(setq bibtex-completion-bibliography '("/c/temp/abbrev0.bib" "/c/temp/crypto.bib"))
(measure-time (bibtex-completion-candidates))
```

I obtained similar loading time before and after the change, around 23.9 seconds.

From here, one possibility that seems appealing to me would be to generalize the idea further by storing the cache data for each bibliography file separately rather than for the whole bibliography. Basically we would simply need to replace the `'global` and `'local` symbols by the bibliography file names in `bibtex-completion-cache`, and make `bibtex-completion-candidates` parse and process entries file per file rather than globally. This would be more efficient if the user modifies one out of several bibliography files or has files that are common to the global and local bibliographies. Note that the cross-references would then also be resolved file per file rather than globally, I don't know if this problematic. Anyway, I can try it if you are interested.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -949,6 +949,19 @@ entry for each BibTeX file that will open that file for editing."
              bib-files)
       bibtex-completion-fallback-options)))
 
+(defun bibtex-completion-find-local-bibliography ()
+  "Return a list of BibTeX files associated with the current file. If the current file is a BibTeX file, return this file. Otherwise, try to use `reftex' to find the associated BibTeX files. If this fails, return `bibtex-completion-bibliography'."
+  (or (and (buffer-file-name)
+           (string= (or (f-ext (buffer-file-name)) "") "bib")
+           (list (buffer-file-name)))
+      (and (buffer-file-name)
+           (require 'reftex-parse nil t)
+           (reftex-locate-bibliography-files
+            (if (fboundp 'TeX-master-directory)
+                (TeX-master-directory)
+              (file-name-directory (buffer-file-name)))))
+      bibtex-completion-bibliography))
+
 (provide 'bibtex-completion)
 
 ;; Local Variables:
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -269,15 +269,11 @@ the directories listed in `bibtex-completion-library-path'."
   :group 'bibtex-completion
   :type 'string)
 
-(defvar bibtex-completion-bibliography-hash nil
-  "The hash of the content of the configured bibliography
-files.  If this hash has not changed since the bibliography was
-last parsed, a cached version of the parsed bibliography will be
-used.")
+(defvar bibtex-completion-cache '((global "") (local ""))
+  "A cache storing the hash of the bibliography content and the corresponding list of candidates, obtained when the bibliography was last parsed. When the current bibliography hash is identical to the cached hash, the cached list of candidates is reused, otherwise the bibliography is reparsed. The global and local bibliographies are cached separately.")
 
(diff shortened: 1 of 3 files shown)


tmalsburg:
Great work, thank you!  I agree with you that one cache per file is cleaner and also more efficient in some cases.  Cross-references are a problem but perhaps it's better to resolve them not during parse time but when an entry is actually used. In this case the resolution of cross-references could be global.

Do you want to amend this PR or shall we merge it as is, and do the one-cache-per-file in a separate PR.  The changes should be transparent to the users, so I wouldn't mind merging what we have to make the local search available.

ericdanan:
Sure feel free to merge this now, I can send another one for the per-file cache.

ericdanan:
One question, about the possibility to resolve crossrefs only when we "actually use" an entry, do you mean that we only need the crossref information when we apply some actions to the selected entries? That is, we don't need it to search or format entries?

tmalsburg:
Merged.  Thanks also to @jagrg!

> One question, about the possibility to resolve crossrefs only when we "actually use" an entry, do you mean that we only need the crossref information when we apply some actions to the selected entries? That is, we don't need it to search or format entries?

Yes, that's what I was thinking but it's actually not going to work.  The year for example is usually retrieved from the cross-referenced entry and we need this for searching.  So we have to resolve crossrefs earlier. Perhaps, we first populate the cache with unresolved entries from all files (one cache per file, as you suggested) and when done resolve all entries such that we have globally resolved entries in the cache.  

One thing to keep in mind: ivy-bibtex caches formatted entries but helm-bibtex caches the alists.

jagrg:
Thank you both. Works great 👍

ericdanan:
> One thing to keep in mind: ivy-bibtex caches formatted entries but helm-bibtex caches the alists.

Yes but this is what makes `ivy-bibtex` slower than `helm-bibtex` and I think it would be better to make the former, like the latter, cache alists and only format filtered candidates. Hence I first implemented and tested this, I am going to send you a PR.

tmalsburg:
To my knowledge, Ivy doesn't support formatting at search time.  That's why we cache the formatted entries.

ericdanan:
Yes there is `ivy-set-display-transformer`, see #140 .

PR #140, opened 2016-09-28
make ivy-bibtex faster by only formatting displayed candidates

ericdanan:
Currently `helm-bibtex` caches unformatted entries and only formats the filtered ones through helm's `filtered-candidates-transformer` mechanism. `ivy-bibtex`, on the other hand, formats all entries before caching them.

This difference makes `ivy-bibtex` much slower than `helm-bibtex` for large bibliographies, and this PR fixes this by making `ivy-bibtex` cache unformatted entries and only format the displayed entries through ivy's `display-transformer` mechanism. There are some differences between helm's and ivy's mechanisms, which we have to take into account, but once we do that the two frontends consistently cache unformatted entries and pass list of keys to their actions, and they are similar in terms of speed. Here are the timings in seconds that I obtain with `crypto.bib` (same methodology as in #137 ):

|      | before | after |
| --- | --- | --- |
| `bibtex-completion-candidates` | 23.8 | 23.8 |
| `helm-bibtex` | 24.0 | 24.0 |
| `ivy-bibtex` | 38.6 | 24.0 |

There are five commits in the PR. The first two are the main ones that actually do the job. The last three are just cleanup that naturally follows from the first two, in my opinion, but I don't know if you will agree so I made separate commits. A detailed explanation follows. Of course let me know if you would like me to change some things.

The first commit makes `ivy-bibtex` cache unformatted entries (like `helm-bibtex`), and then only format the displayed entries through `ivy-set-display-transformer`. Unlike helm, ivy transforms only the displayed candidates and not all the filtered ones, and it transforms them one by one. Also, the transformer function only takes as argument the string to be formatted (that is, the `car` of the candidate), and outputs the formatted string to be displayed. So we have to recover the corresponding alist (the `cdr` of the candidate) from the unformatted string. ivy makes this fast and easy by storing the candidate's index in the candidates list as a text property in the unformatted string. So we simply get the index, fetch the corresponding entry, and use its alist to generate the formatted string. To do this efficiently I separated the formatting code out of the function `bibtex-completion-candidates-formatter` into a new function `bibtex-completion-format-entry`.

Another difference between ivy's and helm's mechanisms is that ivy only transforms the candidates for display but passes the untransformed selected candidate to actions. So `ivy-bibtex` now passes a cons cell (string . alist), instead of (string . key) previously (`helm-bibtex`, on the other hand, passes a list of keys). This makes `bibtex-completion-normalize-candidates` fail to correctly pass keys to actions. The second commit fixes this. But instead of tweaking `bibtex-completion-normalize-candidates` to figure out where to find the keys to pass without knowing which frontend is used, it seemed more natural to me to let each frontend do the job of extracting and passing keys to actions. `helm-bibtex-helmify-action` makes sure that `helm-bibtex` always passes a list of keys. I simply defined an `ivy-bibtex-ivify-actions` macro to make sure that `ivy-bibtex` also passes a similar list (with a single key, since ivy does not support selecting multiple candidates).

As I said, at this point everything works and the remaining commits just do some cleanup:
- The third one moves `bibtex-completion-candidates-formatter` to `helm-bibtex.el`, since only this frontend now uses it (the new function `bibtex-completion-format-candidate`,  which is used by both frontends, remains in `bibtex-completion.el`). In fact I merged this function into `helm-bibtex-candidates-formatter`.
- The fourth one removes the optional `formatter` argument to `bibtex-completion-candidates`, since it is no longer used.
- The fifth one removes `bibtex-completion-normalize-candidates`, which I think is no longer needed since both frontends now consistently pass list of keys to actions. It seems simpler to me to let the generic action functions in `bibtex-completion.el` assume that they systematically receive list of keys, and then let each frontend modify these functions as needed.

Sorry for the lengthy explanations, I hope at least they are clear.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -284,15 +284,12 @@ actually exist."
                   (user-error "BibTeX file %s could not be found." file)))
         (-flatten (list bibtex-completion-bibliography))))
 
-(defun bibtex-completion-candidates (&optional formatter)
+(defun bibtex-completion-candidates ()
   "Reads the BibTeX files and returns a list of conses, one for
 each entry.  The first element of these conses is a string
 containing authors, editors, title, year, type, and key of the
 entry.  This is string is used for matching.  The second element
-is the entry (only the fields listed above) as an alist.
-
-If non-nil, the entries are passed to the function FORMATTER
-before being saved."
+is the entry (only the fields listed above) as an alist."
   ;; Open configured bibliographies in temporary buffer:
   (with-temp-buffer
     (mapc #'insert-file-contents
@@ -312,9 +309,7 @@ before being saved."
                               (s-join " " (-map #'cdr it))) it)
                        entries)))
           (setf (cddr (assoc bibtex-completion-bibliography-type bibtex-completion-cache))
-                (if (functionp formatter)
-                    (funcall formatter entries)
-                  entries)))
+                entries))
         (setf (cadr (assoc bibtex-completion-bibliography-type bibtex-completion-cache))
               bibliography-hash))
       (cddr (assoc bibtex-completion-bibliography-type bibtex-completion-cache)))))
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -522,53 +522,32 @@ values."
                concat (car p))
     nil))
 
-(defun bibtex-completion-normalize-candidate (candidate)
-  "Extract key from CANDIDATE."
-  (cond ((stringp candidate)
-         candidate)
-        ((stringp (cdr-safe candidate))
-         (cdr candidate))))
-
-(defun bibtex-completion-normalize-candidates (candidates)
-  "Normalize CANDIDATES to a list of candidates for each action
-function."
-  (cond
-   ;; single cons cell
-   ((stringp (cdr-safe candidates))
-    (list (cdr candidates)))
-   ((listp candidates)
-    (-map #'bibtex-completion-normalize-candidate candidates))
-   (t
-    (list candidates))))
-
 
-(defun bibtex-completion-open-pdf (candidates)
+(defun bibtex-completion-open-pdf (keys)
   "Open the PDFs associated with the marked entries using the
 function specified in `bibtex-completion-pdf-open-function'.  All paths
 in `bibtex-completion-library-path' are searched.  If there are several
 matching PDFs for an entry, the first is opened."
   (--if-let
       (-flatten
-       (-map 'bibtex-completion-find-pdf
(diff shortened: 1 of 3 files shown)


tmalsburg:
Thank you for the PR and the explanations.  It all looks very clean and well thought-out.  Just one comment:

In `ivy-bibtex-display-transformer`, you use `frame-width`.  If `frame-width` is like `window-width`, I think this won't work under some circumstances because it may give you more characters than actually fit on a line and this would cause unnecessary line breaks or continuation glyphs.  That's why I introduced `helm-bibtex-window-width` which at least takes the size of the fringes into account (but not the font size).  In Emacs 25, there is `window-max-chars-per-line` which is more precise (my first contribution to Emacs), but we can't rely on this function because it's too new. Perhaps move `helm-bibtex-window-width` to bibtex-completion and also use it in ivy-bibtex?

ericdanan:
`frame-width` was used already in `ivy-bibtex-candidates-formatter`. I noticed that `helm-bibtex` was using the more sophisticated `helm-bibtex-window-width` and followed the link in the comment above this function, but I didn't fully understand the issues there. I supposed that the situation was probably simpler with `ivy-bibtex` (maybe because ivy always uses the minibuffer to display candidates) and this was probably why `frame-width` was simply used there.

Anyway, I don't know what the problematic circumstances may be, but for me `ivy-bibtex` perfectly fits candidates to the minibuffer width (and so does `helm-bibtex`, whether in full frame or not). But if I use `helm-bibtex-window-width` instead of `frame-width` in `ivy-bibtex`, then it no longer works when I have two windows side by side in the frame: the candidates in the minibuffer then only use the width of the window from which I called `ivy-bibtex` instead of the full minibuffer width.

So I don't know how to do better than `frame-width`...

tmalsburg:
Ok, then lets stick with frame-width.

ericdanan:
I would say it is ready to be merged. I have tested `ivy-bibtex` and `helm-bibtex` and tried the different actions, everything worked for me. The only problem I can think of is if users have customized the actions for `ivy-bibtex`, then it will stop working for them until they update their setup to use the "ivified" actions. I don't see how we can avoid this. Should we mention it in the `README`?

tmalsburg:
Re ivyfied actions: Perhaps add a news item in `README.org` and one sentences saying that they have to use `ivy-bibtex-...` instead of `bibtex-completion-...` for actions.

tmalsburg:
I changed the README and the added an item in the news section of ivy-bibtex.el.  Thank you for this PR, great work!

PR #142, opened 2016-10-05
Cache improvements

ericdanan:
Hi,

There are two commits here. They do different things but both related to the way bibtex-completion caches entries.

The first commit implements the per-file cache instead of the current global / local cache. As discussed is #137 this is cleaner, and more efficient in case only part of the bibliography files are updated or some files are shared between the global and local bibliographies, but we have to be careful because cross-references must be resolved globally. So the strategy is that the cache is now an alist of elements of the form `(FILE HASH ENTRIES)` and `bibtex-completion-candidates` now works as follows:

1. We first load each `FILE` individually and compute the `HASH`. If it has changed with respect to the cached `HASH` we parse the file and update its `HASH` and `ENTRIES`.
2. Once this is done for all bibliography files, we make a hash-table of all entries to resolve cross-references globally and we prepare the entries (the prepared entry is the cons cell `(SEARCH-STRING . ENTRY-ALIST)` that is passed to the frontend). But rather than repreparing all entries whenever some files were reparsed, it is more efficient to do this only for the entries from the reparsed files as well as the entries with a crossref field from the other files, and simply reuse the prepared entry from the cache for the entries from non-reparsed files without a crossref field. So we do what is needed for each `FILE` and update its `ENTRIES` with the prepared entries. When we update `ENTRIES`, we don't replace the "raw" entries (as parsed in step 1) with the prepared entries, but rather store the latter alongside the former. So each element of `ENTRIES` is now a cons cell `(RAW-ENTRY . PREPARED-ENTRY)`. This way, if entry A from a non-reparsed file crossreferences entry B from a reparsed file, we can still start from the raw entry A to resolve the crossref.
3. Finally, we collect and return the prepared entries from `ENTRIES` for each `FILE`.

This change should be transparent to the user. I measured loading time as in #137 and did not see any impact when we first call `helm-bibtex` or `ivy-bibtex`, or if we call them with an up-to-date cache. Of course we gain a lot of time if we update only one of several bibliography files. I tested both `helm-bibtex` and `ivy-bibtex` with global and local bibliographies, everything seems to work, but of course you may want to do more testing.

The second commit makes a simpler change: checking for pdf and notes only for filtered candidates. That is to say, this is no longer done in `bibtex-completion-prepare-entry` but rather in `bibtex-completion-format-entry`. This significantly reduces loading time for large bibliographies: again as in #137 , it goes down from `21.5` to `10.5` seconds (note that the `21.5` was `23.9` in #137 but this seems to be because I updated emacs from version `24.5` to `25.1`). Then once the bibliography is loaded and I type a search pattern, I don't notice that the candidates take longer to update than before (although I don't know how to properly measure this): `ivy-bibtex` updates instantly whereas `helm-bibtex` has a small delay, but the delay does not seem longer. This makes sense to me: if we gain `11` seconds by no longer checking pdf and notes for about `40000` entries, then on the other hand we should proportionally lose about `0.1` second everytime the search pattern changes to do so for the `500` filtered candidates only, which is not noticeable. I suppose the small noticeable delay with `helm-bibtex` could come from the fact that it always formats `500` filtered entries whereas `ivy-bibtex` only formats the (usually less than `10`) displayed entries, although I did not check that.

Another consequence of this change is that the pdf and notes information is always up-to-date, since it is now read "in real time" rather than from the cache.

Again you may want to test this by yourself and see if you feel that it makes `helm-bibtex` less responsive.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -269,11 +269,8 @@ the directories listed in `bibtex-completion-library-path'."
   :group 'bibtex-completion
   :type 'string)
 
-(defvar bibtex-completion-cache '((global "") (local ""))
-  "A cache storing the hash of the bibliography content and the corresponding list of candidates, obtained when the bibliography was last parsed. When the current bibliography hash is identical to the cached hash, the cached list of candidates is reused, otherwise the bibliography is reparsed. The global and local bibliographies are cached separately.")
-
-(defvar bibtex-completion-bibliography-type 'global
-  "Whether to use the global or local bibliography.")
+(defvar bibtex-completion-cache nil
+  "A cache storing the hash of the bibliography content and the corresponding list of entries, for each bibliography file, obtained when the bibliography was last parsed. When the current bibliography hash is identical to the cached hash, the cached list of candidates is reused, otherwise the bibliography file is reparsed.")
 
 
 (defun bibtex-completion-init ()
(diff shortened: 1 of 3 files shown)


tmalsburg:
Thank you for this PR.  I will have a closer look at the code later, but for now I have some quick questions:

1. Does the new code also store a resolved version for entries that do not have a crossref field?  I'm asking because storing resolved and raw entries may increase memory consumption which is already rather high for bibliographies like crypto.bib.

2. How does the code handle duplicates during crossref resolution, i.e., the case where several bibtex files contain an entry with a given name?  Background: Some people have one big bibtex file and then small bibtex files for their research papers.  The small files are usually subsets of the largefile but may contain outdates versions of, for example, in-press papers.

3. More a comment than a question: One effect of the second commit is that it's not anymore possible to search for entries that have notes or PDFs.  Currently this can be done by entering the notes or PDF symbol.  There is clearly a UI problem with this current solution because it's difficult to enter these symbols, but perhaps it would be better to fix the UI issue instead of giving up this feature entirely.

ericdanan:
1. Yes it stores both versions for all entries, so that we can directly reuse the candidate for entries from non-reparsed files without crossref field. Note that the resolved version does not only resolve crossrefs, it also removes / formats some fields and generates the search string. If we don't store this, we have to redo the work each time we call `bibtex-completion-candidates`.  Should I do something to measure the increase in memory consumption? I'm not sure what the problem is. If it is that the computer could be frozen or sluggish until `bibtex-completion-candidates` finishes, it is not the case for me (I can use Firefox normally, for instance).

2. The code doesn't do anything about this, so there will be multiple entries with the same key (is this not currently the case as well?). It seems easy to remove duplicate keys when we return the list of candidates, if we can assume that the first instance is the one to keep (i.e. files listed earlier in `bibtex-completion-bibliography` have priority). I don't know if this would slow down things. I'm not sure what happens in the entry hash-table when we have several entries with the same key, perhaps we would need to remove duplicates here as well.

3. I personally don't mind because although it is useful to quickly see if an entry has a pdf or notes file, I never want to explicitly restrict the search to such entries. But if this can be useful to some users, then there is a tradeoff between loading faster and having this search possibility. Perhaps then we could introduce a custom variable to let each user choose what they want. Or perhaps we could add some key bindings to `helm-bibtex` / `ivy-bibtex` to call some functions that would check all candidates for pdf / notes and filter / unfilter them.

tmalsburg:
1. If I remember correctly the original purpose of creating stripped down entries was to save memory and to speed things up. (Perhaps that was an instance of premature optimization.)  So since we're storing the raw and complete entries anyway, there is probably no need for the stripped down entries as they primarily duplicate information.  I'd say we remove the stripped entries and leave it to `bibtex-completion-make-candidate` to ignore the fields that are not relevant for the search string.

Let's ignore 2. for now. (You are right that this is also an issue in the current code.)

3. The slowdown is more than 100ms on my system (Intel i5) and it's fairly noticeable.  I'm not using the file field to reference PDFs but the scheme bibtex-key.pdf.  The latter is faster and if people use the file field the slowdown will be even bigger.  I can live without the capability to search for entries with PDFs and notes but since you're introducing a more fine-grained and hence more efficient caching system, wouldn't it make sense to search for notes and PDFs at parse time?  Personally, I don't mind waiting a little longer when I call helm-bibtex for the first time.

ericdanan:
1. Again I don't understand what the memory issue is about. If I understand correctly, you suggest that instead of storing `(RAW-ENTRY SEARCH-STRING . PREPARED-ENTRY)` and returning `(SEARCH-STRING . PREPARED-ENTRY)`, we could simply store and return `(SEARCH-STRING . RAW-ENTRY)`. Am I correct? If we do this I think there will be a problem when `helm-bibtex` / `ivy-bibtex` formats the entry for display: if the entry has a crossref field then `RAW-ENTRY` won't have the crossref information (e.g. the year).

2. OK.

3. OK then I should simply discard the second commit. But I'm not sure what's the correct way to do it: should I do a hard reset in my fork's `cache` branch?

tmalsburg:
Re 1: Yes, that's what I was suggesting.  You are right about the crossrefs.  Could we generate the prepared entries on the fly at search time or would that be too slow?  All we have to do is to check whether an entry has a crossref field (most don't in which case we're done) and then look up the crossreffed entry in the hashtable and copy some stuff.  Shouldn't be too bad.

Re 3: Hard reset and forced push should work but I haven't tried it.

ericdanan:
If I understand correctly, your suggestion is basically to cache only the raw entries, with pdf and notes checked but crossrefs unresolved. And then just before returning the candidates `bibtex-completion-candidates` would resolve the crossrefs without caching this information. But then the crossrefs would have to be resolved every time `bibtex-completion-candidates` is called, even if no bibliography file was changed. This takes about `6` seconds for me with `crypto.bib`, so I think it would better be avoided.

I thought of another solution. To avoid caching redundant information, we can cache only the candidates as before (ie conses `(SEARCH-STRING . RESOLVED-ENTRY)`, but in `RESOLVED-ENTRY` we insert an empty field `("" . "")` between the entry's own fields and the fields imported from the crossrefed entry. This empty field has no impact on entry matching and formatting, but it allows to discard the crossref info if we later have to resolve the entry again. Here is how `bibtex-completion-candidates` now works in more detail:

1. Check each bibliography file and reparse it if needed. When reparsing a file, each time we read an entry we prepare it right away (essentially remove unwanted fields and check for pdf / notes). We only cache these prepared entries.

2. If some files were reparsed then we resolve crossrefs:
    - For a reparsed file, if an entry has a crossref field then we add the fields from the crossrefed entry to it but we add an empty field beforehand. Then we make the candidate for this entry (generate the search string and cons it with the entry). If an entry has no crossref field we simply make the candidate. Then in the cache we replace the entries from step 1 with the candidates.
    - For a non-reparsed file, if an entry has a crossref field we resolve it and make the candidate as above, but we first cut the entry at the empty field so as to discard the old crossref information. If an entry has no crossref field we simply reuse the cached candidate. As above in the cache we replace the entries from step 1 with the candidates.

3. Finally we return the cached candidates.

It seems to me that this way we avoid caching redundant information. The cache has the same structure as before, there is only the additional empty field in the entries with a crossref field, which is not a big deal I suppose.

I made a hard reset to get rid of the second commit (check pdf / notes), then amended the first commit (per-file cache) to make it work as I just described, and finally made a forced push. Let me know if you have any issue with the forced push or the amended commit.

tmalsburg:
> If I understand correctly, your suggestion is basically to cache only the raw entries, with pdf and notes checked but crossrefs unresolved. 
Yes, that was the idea.  Just out of curiosity, do you know how were those 6 seconds were spend?  Building the hashtable or doing the actual resolution or both more or less equally?

The forced push hasn't caused any problems so far.  I find the current solution really neat but will test it a little before merging.

ericdanan:
Right now `bibtex-completion-resolve-crossrefs` takes `5.6` seconds. Out of this, `bibtex-completion-make-entry-hash` takes only `0.7` seconds, so most of the time is spent on actually resolving the crossrefs and making the candidates.

tmalsburg:
Works beautifully.  Thank you very much for this PR. Very much appreciated.

jagrg:
This is working very well. Thanks for the PR! The only problem I've had so far is the list of bibliographies are now inverted. For example, with `(setq bibtex-completion-bibliography '("~/a.bib" "~/b.bib"))` defined, the first candidates start from b.bib. I think it makes sense to start with a.bib instead, as it did before, in which case we probably need `(nreverse bibtex-completion-bibliography)` somewhere.

tmalsburg:
I'm not sure whether it is the most logical thing to start with entries from a.bib (because within a .bib file the entries are reversed) but for the sake of consistency we should stick with the old ordering.  Fixed in 072aafb00979b5bd8ea06f4585b5cd21aefa67eb

ericdanan:
It's fine with me either way, but was the old behavior really to put candidates from a.bib first? I may have missed something here, but it seems to me that the old code was doing the following:
1. Insert the content of a.bib and then b.bib in a temp buffer
2. Parse the entries in the order in which they appear in the temp buffer
3. Return the whole list of entries
In which case the first candidates should be the last entry from b.bib and the last candidate should be the first entry from a.bib. In fact I tried to replicate this, and the order of entries did not change for me.

tmalsburg:
I always used just one bib file, but I think Eric is right.

jagrg:
I've used two bib files for a while and AFAICT the order has always been:

a.bib (last entries first)
b.bib (first entries first)

In any case, I'm fine with switching the order if that's a better choice. I suppose we can still reorder the list manually.

jagrg:
There's something strange happening. Here's how to reproduce it.

1. Define two bib files, e.g.: (setq bibtex-completion-bibliography '("~/a.bib" "~/b.bib"))
2. Restart Emacs
3. Call "M-x helm-bibtex"
4. Check the value of `bibtex-completion-bibliography`
5. b.bib is missing.

ericdanan:
OK my mistake, Jonathan is right actually, I just checked with master branch before pulling this PR. The reason is that `insert-file-contents` inserts the file contents after point. So with the old code, a.bib was inserted after point and then b.bib was inserted after point and hence before a.bib. So the last commit indeed preserves the former ordering of candidates.

ericdanan:
I can reproduce this after the last commit. The `nreverse` docstring says it may destructively modify the list. Moving the `nreverse` just before `-flatten` seems to resolve the issue, see #145.

jagrg:
Thanks, #145 fixed it.

PR #143, opened 2016-10-06
Added bibtex-completion-open-any

plantarum:
I added a new default action, which rolls open pdf and open url/doi together. My usual workflow is to open the local pdf if I have one, and if not to immediately try to open a url or doi. Having separate manual actions to accomplish this is inconvenient.

In the pull request, my new function is placed as default, which is presumptuous of me. If the idea is sound but you don't want it as default, that makes sense. Also, I have used `dolist` only, rather than the fancy functional stuff in dash.el. Again, if the idea is sound, maybe you'd prefer a dash-based solution.

So, if any of this looks sensible, I'm happy to make any alterations necessary to suit you. And if you don't like it, I'll deal with it locally for my own use.

Best,

Tyler

### DIFF
--- a/README.org
+++ b/README.org
@@ -9,6 +9,7 @@ Helm-bibtex and ivy-bibtex allow you to search and manage your BibTeX bibliograp
 Ivy-bibtex is experimental at this time, lacks some features, and may be noticeably slower than helm-bibtex when used with larger bibliographies.
 
 * News
+- 2016-10-07: Added bibtex-completion-open-any, made it the default action for helm-bibtex and ivy-bibtex
 - 2016-09-29: Performance improvements in ivy-bibtex.  *Note:* If you changed your default action in ivy-bibtex, you have to rename the action, e.g. from ~bibtex-completion-insert-key~ to ~ivy-bibtex-insert-key~.  See [[#change-the-default-action][here]] for details.
 - 2016-09-20: Added fallback options to ivy frontend.
 - 2016-04-18: Improved support for Mendely/Jabref/Zotero way of referencing PDFs.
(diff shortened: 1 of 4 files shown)


tmalsburg:
I like that and almost wonder if we can drop the old action for opening PDFs.  Also, good old-fashioned `do-list` is fine.  As you can see, my code is pretty eclectic and happily mixes cl-loop, dash, and standard elisp looping facilities.

plantarum:
Dropping the old pdf opener would be fine with me. I'll do so and tidy up the documentation when I get a moment.

tmalsburg:
Let's replace `helm-bibtex-open-pdf` by `helm-bibtex-open-any` in the list of actions but keep the old function around for people who prefer to use that.  For consistency, the same changes should be made for ivy-bibtex.

plantarum:
Do you want to keep helm-bibtex-open-url-or-doi in the list?

tmalsburg:
We need it, without it it would be impossible to open the URL when the article has a PDF.

plantarum:
Of course, that makes sense! In that case, I think all the discussed changes, including updating ivy-bibtex, are in the pull request. I haven't actually used the ivy-bibtex interface, but I don't see any other obvious things to change in that file.

tmalsburg:
Merged. Thank you!

ericdanan:
In the code of `bibtex-completion-open-any`, wouldn't be simpler to replace the whole `(let* ...)` block copied from `bibtex-completion-open-url-or-doi` by just `(bibtex-completion-open-url-or-doi (list key))`?

plantarum:
Looks like it does, and much simpler that way! My code is obviously silly now that you point it out.

While looking at that, I think the form `(message "No URL or DOI found for this entry: %s" key)` in `bibtex-completion-open-url-or-doi` is in the wrong spot - it should be the `else` clause for the `(if doi ...)` form, rather than a *second* form in the `else` clause of the `(if url ...)` form. As is, it always complains when a URL is not found, even if a DOI *is* found.

PR #148, opened 2016-10-16
New action to add a PDF to the library

ericdanan:
The PDF can be added either from an open buffer or a file.


tmalsburg:
Merged.  Good idea!

PR #149, opened 2016-10-17
update readme

ericdanan:
Main changes:
- documented the local bibliography search feature, and the helmify / ivify mechanism to create new actions
- added ivy-bibtex specific documentation in some places where it was missing, and removed the "ivy is experimental..." sentence and the one saying that crossref import is specific to helm-bibtex, as I think both frontends now offer more or less the same features and performances
- updated some points that seemed out of date to me

Overall there are a lot of small changes, so I gave making an exhaustive list or making a separate commit for each one. At some places I was led to rewrite things a bit to put together the helm-bibtex and ivy-bibtex specific documentation more efficiently, I hope, but this is necessarily subjective so feel free to rewrite or ask me to do so.


tmalsburg:
This is great, thank you very much!

PR #150, opened 2016-10-18
Fix bibtex-completion-find-pdf-in-field for simple paths.

mwlodarczak:
bibtex-completion-find-pdf-in-field stopped working for simple paths of the form 
> File = {/path/to/article.pdf}

This simple fix solves this problem.


tmalsburg:
Merged. Thank you.

PR #154, opened 2016-10-27
No quotes are needed when creating new helm actions

Titan-C:
I'm quite a ELisp beginner. But after trying to get a new action (open pdf and notes at the same time) I noticed this quotes are not needed.


tmalsburg:
True, but while you are at it, could you also remove the equivalent quotes for ivify? Thank you!

Titan-C:
Removed the ivify quotes.

tmalsburg:
Thank you, Óscar!

tmalsburg:
And have fun hacking elisp :-)

PR #156, opened 2016-10-30
don't change point when searching bibtex entries

bastibl:
In my bibliography notes file, I have entries similar to this one
```
** 2003 - Software Defined Radio: Enabling Technologies
   :PROPERTIES:
   :Custom_ID: tuttlebee2003software
   :AUTHOR:   Tuttlebee
   :YEAR:     2003
   :END:
   cite:tuttlebee2003software
```
When the point is on ```cite:...```, opening the context menu moves the point to the end of the ```Custom_ID``` line. Subsequent actions like opening the bibtex entry, therefore, fail as the point is no longer on a bibtex key.

Debugging showed that the ```with-temp-buffer``` function of ```bibtex-completion-get-entry1``` does not save the point and should be wrapped in a ```save-excursion```.

(I have no idea about lisp, so there might be a much nicer way to do this)

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -430,15 +430,16 @@ appended to the requested entry."
     (bibtex-completion-remove-duplicated-fields (append entry crossref))))
 
 (defun bibtex-completion-get-entry1 (entry-key &optional do-not-find-pdf)
-  (with-temp-buffer
-    (mapc #'insert-file-contents
-          (-flatten (list bibtex-completion-bibliography)))
-    (goto-char (point-min))
-    (re-search-forward (concat "^@\\(" parsebib--bibtex-identifier
-                               "\\)[[:space:]]*[\(\{][[:space:]]*"
-                               (regexp-quote entry-key) "[[:space:]]*,"))
-    (let ((entry-type (match-string 1)))
-      (reverse (bibtex-completion-prepare-entry (parsebib-read-entry entry-type) nil do-not-find-pdf)))))
+  (save-excursion
+    (with-temp-buffer
+      (mapc #'insert-file-contents
+            (-flatten (list bibtex-completion-bibliography)))
+      (goto-char (point-min))
+      (re-search-forward (concat "^@\\(" parsebib--bibtex-identifier
+                                "\\)[[:space:]]*[\(\{][[:space:]]*"
+                                (regexp-quote entry-key) "[[:space:]]*,"))
+      (let ((entry-type (match-string 1)))
+        (reverse (bibtex-completion-prepare-entry (parsebib-read-entry entry-type) nil do-not-find-pdf))))))
 
 (defun bibtex-completion-find-pdf-in-field (key-or-entry)
   "Returns the path of the PDF specified in the field


tmalsburg:
I think this is an issue in org-ref not helm-bibtex.  helm-bibtex doesn't define those context menus and I actually find it surprising that this patch fixed the issue.  The code in `bibtex-completion-get-entry1` doesn't mess with the org buffer and the `save-excursion` should therefore not be necessary.  Perhaps, I'm missing something?

bastibl:
AFAIS, it doesn't mess with the buffer, but moves the point away from the bibtex key. When org-ref is actually called, the point in no longer on a bibtex key and it org-ref fails.

tmalsburg:
> but moves the point away from the bibtex key. 

Where in the code is this happening?  AFAICS, the point is moved but only in the temporary buffer.

bastibl:
Hmm, I hoped that would be obvious :-/ To, at least, show that it's actually the case, I patched ```bibtex-completion-get-entry```

```
(defun bibtex-completion-get-entry (entry-key)
  "Given a BibTeX key this function scans all bibliographies
listed in `bibtex-completion-bibliography' and returns an alist of the
record with that key.  Fields from crossreferenced entries are
appended to the requested entry."
  (let* ((p1 (point))
         (entry (bibtex-completion-get-entry1 entry-key))
         (p2 (point))
         (crossref (bibtex-completion-get-value "crossref" entry))
         (crossref (when crossref (bibtex-completion-get-entry1 crossref))))
    (print p1)
    (print p2)
    (bibtex-completion-remove-duplicated-fields (append entry crossref))))
```

Without ```save-excursion``` it changes in some files.

tmalsburg:
No need to be rude.  Also, I cannot reproduce this.  I don't have org-ref and when I execute `bibtex-completion-get-entry` in a buffer the point is exactly in the same place afterwards.  Please provide a reproducible example that does not require org-ref.  If you can't make such an example, you should submit a patch to org-ref.

bastibl:
Sorry, but I have no idea what you considered rude in my last reply. Maybe that I hoped that the bug would be obvious to someone who's more into lisp. I don't even have an idea what this function is doing and hoped that I could point you to the right function and help debugging this.

Also, I don't see a relation to org-ref. I thought printing the point before and after the call to ```bibtex-completion-get-entry1``` showed that this function actually changes the point in some (rare) cases. Looks like this was not the correct way to debug.

tmalsburg:
> I have no idea what you considered rude in my last reply.

You said that the answer to my question should have been obvious, but perhaps I misunderstood you. In that case, I'm  sorry.

Anyway, org-ref sometimes modifies parts of helm-bibtex and in the past that has been a source of bugs.  As far as I can see there is no code in the `bibtex-completion-get-entry1` that should move the point (except in the temporary buffer) and the `save-excursion` should therefore be unnecessary.  I'm happy to apply a fix, but I first want to understand what exactly is going on here. Since I can't reproduce the problem, I suspect the root of it lies outside of helm-bibtex.

bastibl:
> You said that the answer to my question should have been obvious

I wrote that I hoped it would be obvious not think that it should be obvious...

The problem is that this is my notes file. While assembling the metadata for the bibtex key. ```bibtex-completion-prepare-entry``` open the notes file (i.e., the current file and searches for its Custom_ID: <bibtexkey>. This changes the point of the current file, leading to the behavior described above.

So ```bibtex-completion-prepare-entry``` changes the point in the notes file, which was in that particular case my current file. Adding ```save-excursion``` there fixes my problem.

tmalsburg:
I see, that makes sense.  I tested with an org file that was not the notes file.  I agree that a save-excursion should be added, but it's better to do it in `bibtex-completion-prepare-entry`.  Fixed in
8585af02c4afdf18b2069628085c2a5dcc272c9d.  Thanks for reporting this issue.

bastibl:
Great, thanks!

tmalsburg:
I was actually also experiencing strange issues that appeared to be unrelated but were fixed but this.

PR #157, opened 2016-11-02
Handle the case where the pdf-field exists but is empty.

lixinchin:
In the rare case when the `bibtex-completion-pdf-field' field exists for
a particular bibtex entry, but the file(s) it points to are just the empty
string, `bibtex-completion-find-pdf-in-field' should return `nil' instead
of "".


tmalsburg:
Good catch. Thanks for the PR.

PR #158, opened 2016-11-05
make candidate format customizable

ericdanan:
Created a new custom variable `bibtex-completion-display-formats` and
modified the function `bibtex-completion-format-entry` accordingly. The
former variable is an alist of format strings for displaying entries in the results list. The key of each element of this list is either a BibTeX entry type (in which case the format string applies to entries of this type only) or t (in which case the format string applies to all other entry types). The value is the format string.

In the format string, expressions like \"${author}\", \"${title}\", etc,
are expanded to the value of the corresponding field. An expression like
\"${author:N}\" is further truncated to a width of N characters, whereas
an expression like \"${title:-N}\" is truncated to a width equal to the
results window's width minus N characters. Three special fields are
available: \"=type=\" holds the BibTeX entry type, \"=has-pdf=\" holds
`bibtex-completion-pdf-symbol' if the entry has a pdf file, and
\"=has-notes=\" holds `bibtex-completion-notes-symbol' if the entry has
a notes file.

The default format is unchanged.


tmalsburg:
This is really great. It would be good if we could write something like `${title:*}` to specify that any remaining characters should be used for that field.  This way users don't have to do the math to find out how much space is left.  The notation `${title:-53}` could be dropped then.  Is there a simple way to achieve this?

ericdanan:
I implemented this in the last commit, it seems to work as expected, but I didn't try with a large bibliography to see if it makes helm-bibtex slower (ivy-bibtex is fine but helm-bibtex formats 500 entries at so mignt make a difference)

ericdanan:
I gave it a try with `helm-bibtex` and crypto.bib. It seems OK to me (`helm-bibtex` is less responsive than `ivy-bibtex` but I don't see a difference after the last commit), but perhaps you could test it as well to make your own opinion (I suppose you could use any bibliography with 500+ entries instead of crypto.bib)?

tmalsburg:
Great!  There is a slowdown but it's small, so we could definitely merge this.

One thing that's slightly suboptimal is that we're computing the remaining width again and again for each entry when it would be sufficient to do it once for each entry type.  Specifically, we could calculate the remaining width in `helm-bibtex-candidates-formatter` and just insert that width into the format string:

``` elisp
(defun helm-bibtex-candidates-formatter (candidates _)
  (cl-loop
    for display-format = (bibtex-completion-make-format-strings)
    ...
    collect (let ((bibtex-completion-display-format display-format))
  	          (cons (bibtex-completion-format-entry entry width) entry-key))))
```

`bibtex-completion-make-format-strings` would just replace the `*` with a concrete value.  Then `bibtex-completion-format-entry` doesn't need to know about variable field sizes and only interprets `${fieldname:N}` with `N` being >= 0.

While this would work for helm-bibtex, I don't know how we would implement this for ivy-bibtex, because `ivy-bibtex-display-transformer` takes only one candidate at a time. Any ideas?

By the way, don't feel obliged to work on the above proposal. We could also merge what we have and do optimizations later.  Thanks again for working on this!

tmalsburg:
@jkitchin, I'd like to make you aware of this PR.  We're soon going to change how entries are formatted.  There will be a very expressive facility for customizing the format on a per-entry-type basis (basically an enriched version of s-format's format strings).  I think this might make some of your modifications unnecessary.  However, this change once merged may also break your current integration of helm-bibtex.

ericdanan:
Here it is with the last commit. I used a slightly different approach that works with both `helm-bibtex' and `ivy-bibtex': I added a variable `bibtex-completion-display-formats-internal' to store the display formats together with the "used width" of each format string (i.e. the width that must be subtracted from the window width to obtain the width of the `:*` field). This variable is set internally by `bibtex-completion-init`. Then `bibtex-completion-format-entry` reads both the format string ant its used width from this variable and hence don't need to compute the `:*` width any longer.

jkitchin:
Thanks for the heads up. It sounds like a great feature.

tmalsburg:
Many thanks, Eric.  I already love this feature.  Here's my format string (work in progress):

``` elisp
      '(("inproceedings" . "${author:36} | ${title:*} | ${booktitle:40} | ${year:4} ${=has-pdf=:1}${=has-note=:1} ${=type=:7}")
        ("article"       . "${author:36} | ${title:*} | ${journal:40} | ${year:4} ${=has-pdf=:1}${=has-note=:1} ${=type=:7}")
        ("phdthesis"     . "${author:36} | ${title:*} | ${year:4} ${=has-pdf=:1}${=has-note=:1} ${=type=:7}")
        (t               . "${author:36} | ${title:*} | ${year:4} ${=has-pdf=:1}${=has-note=:1} ${=type=:7}")))
```

ericdanan:
You're welcome, glad you like it. I'm trying this at the moment:

```
	'((article
	   . "${=has-pdf=:1}${=has-note=:1} ${=type=:3} ${year:4} ${author:36}  ${title:*}  ${journal:40}")
	  (inbook
	   . "${=has-pdf=:1}${=has-note=:1} ${=type=:3} ${year:4} ${author:36}  ${title:*}  Chapter ${chapter:32}")
	  (incollection
	   . "${=has-pdf=:1}${=has-note=:1} ${=type=:3} ${year:4} ${author:36}  ${title:*}  ${booktitle:40}")
	  (inproceedings
	   . "${=has-pdf=:1}${=has-note=:1} ${=type=:3} ${year:4} ${author:36}  ${title:*}  ${booktitle:40}")
	  (t
	   . "${=has-pdf=:1}${=has-note=:1} ${=type=:3} ${year:4} ${author:36}  ${title:*}")))
```

tmalsburg:
Nice, I like that layout.  Just added some documentation and news items.

ericdanan:
Great. Perhaps it's worth mentioning that for that layout to work, one has to add `journal`, `booktitle` and `chapter` to `bibtex-completion-additional-search-fields`?

tmalsburg:
Done, thanks.

Issue #133, opened 2016-09-11
bibtex-completion-show-entry does not expand abbreviations defined in String entry


sewkokot:
Assume in lit.bib file I have this entry:

@Article{neuenhofer97jse,
  author =       {Neuenhofer, A. and Filippou, F. C.},
  title =        {Evaluation of nonlinear frame finite element models},
  journal =      jse,
  year =         1997,
  volume =   123,
  number =   7,
  number =   {958-966}}

and jse abbreviation is defined in other file journals.bib
@String{jse = {Journal of Structural Engineering}}

in bibtex-completion-bibliography I have both bib files. However after [f4] "Insert reference" the abbrev is not expanded:
- Neuenhofer, A., & Filippou, F. C. (1998). Evaluation of nonlinear
  frame finite-element models. jse, 124(6), 704–711.

Any idea how to fix it?

[Update: I meant "Insert reference" and not "Show entry"]

jagrg:
That's a good point. You need `bibtex-expand-strings` set to t, but I think `bibtex-completion-apa-format-reference` would need to be changed as well. Here's a short version of how that might work:

```elisp
(defun bibtex-expand-journal (key)
  (let* ((key (bibtex-completion-normalize-candidate key))
	 (entry (bibtex-completion-get-entry key))
	 (ref (pcase (downcase (bibtex-completion-get-value "=type=" entry))
		("article"
		 (s-format
		  "${author} (${year}). ${title}. ${journal}, ${volume}(${number}), ${pages}.${doi}"
		  'bibtex-completion-apa-get-value entry)))))
    (with-temp-buffer
      (mapcar (lambda (x)
		(insert-file-contents x))
	      bibtex-completion-bibliography)
      (goto-char (point-min))
      (re-search-forward
       (concat "^@\\(" parsebib--bibtex-identifier
	       "\\)[[:space:]]*[\(\{][[:space:]]*"
	       (regexp-quote key) "[[:space:]]*,") nil t)
      (bibtex-beginning-of-entry)
      (let* ((journal (bibtex-text-in-string
		       (bibtex-search-forward-field "journal") t))
	     (s (replace-regexp-in-string (cdr (assoc "journal" entry))
					  journal ref))
	     (s (replace-regexp-in-string "\\([.?!]\\)\\." "\\1" s)))
	s))))
```

With the above, the journal should expand:

`(bibtex-expand-journal "neuenhofer97jse")`

sewkokot:
[...]

PR #161, opened 2016-11-16
support for handling bibtex' @string

malb:
This PR adds initial support for `@string` cf. #133 

It probably needs some refinement as my elisp isn't that great. Also, parsebib does not support bibtex string concatenation so entries like

```
@string{pkc14name =             "PKC~2014" # pkcname_v2}
```

found in [cryptobib](https://cryptobib.di.ens.fr) won't work.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -283,6 +283,12 @@ In the format string, expressions like \"${author:36}\", \"${title:*}\", etc, ar
 (defvar bibtex-completion-cache nil
   "A cache storing the hash of the bibliography content and the corresponding list of entries, for each bibliography file, obtained when the bibliography was last parsed. When the current bibliography hash is identical to the cached hash, the cached list of candidates is reused, otherwise the bibliography file is reparsed.")
 
+(defvar bibtex-completion-string-cache nil
+  "A cache storing bibtex strings, for each bibliography file, obtained when the bibliography was last parsed.")
+
+(defvar bibtex-completion-string-hash-table nil
+  "A hash table used for string replacements.")
+
 
 (defun bibtex-completion-init ()
   "Checks that the files and directories specified by the user
@@ -317,7 +323,42 @@ actually exist. Also sets `bibtex-completion-display-formats-internal'."
                    (or files
                        (-flatten (list bibtex-completion-bibliography)))))
          bibtex-completion-cache)))
-    
+
+(defun bibtex-completion-clear-string-cache (&optional files)
+  "Clears FILES from cache. If FILES is omitted, all files in `bibtex-completion-bibliography' are cleared."
+  (setq bibtex-completion-string-cache
+        (cl-remove-if
+         (lambda (x)
+           (member (car x)
+                   (or files
+                       (-flatten (list bibtex-completion-bibliography)))))
+         bibtex-completion-string-cache)))
+
+(defun bibtex-completion-parse-strings ()
+  "Parse the BibTeX strings listed in the current buffer and
+return a list of entries in the order in which they appeared in
+the BibTeX file."
+  (goto-char (point-min))
+  (let  ((strings (cl-loop
+                   for entry-type = (parsebib-find-next-item)
+                   while entry-type
+                   if (string= (downcase entry-type) "string")
+                   collect (parsebib-read-string))))
+    (-filter (lambda (x) x) strings)))
+
+(defun bibtex-completion--string-hash-table (files)
+  "Construct a hash table for string replacements found in FILES"
+  (cl-loop
+   with entries =
+   (cl-loop
+    for file in files
+    for entries = (cddr (assoc file bibtex-completion-string-cache))
+    append entries)
+   with ht = (make-hash-table :test #'equal :size (length entries))
+   for entry in entries
+   do (puthash (car entry) (gethash (cdr entry) ht (bibtex-completion-clean-string (cdr entry))) ht)
+   finally return ht))
+
 (defun bibtex-completion-candidates ()
   "Reads the BibTeX files and returns a list of conses, one for
 each entry.  The first element of these conses is a string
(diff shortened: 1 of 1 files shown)


tmalsburg:
Hi @malb, thanks for making this PR and sorry for the slow response. I'm busy with teaching. I will send some comments later this week (hopefully).

malb:
No worries, I can relate :)

tmalsburg:
Hi, this is a useful feature but I still think that string resolution should be handled by the API used of parsing BibTeX.  Would you mind creating an issue in [parsebib](https://github.com/joostkremers/parsebib/issues)?

tmalsburg:
Perhaps we should consider switching to bibtex.el for parsing? We'd lose a dependency and according to Kremers, bibtex.el is more sophisticated.

malb:
I've updated this PR to use https://github.com/joostkremers/parsebib/pull/6 It is indeed quite a bit cleaner.

tmalsburg:
Great. I'll merge it once your PR for parsebib is closed. 

On March 8, 2017 10:48:32 AM EST, Martin Albrecht <notifications@github.com> wrote:
>I've updated this PR to use
>https://github.com/joostkremers/parsebib/pull/6 It is indeed quite a
>bit cleaner.
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/tmalsburg/helm-bibtex/pull/161#issuecomment-285078174

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

joostkremers:
I closed https://github.com/joostkremers/parsebib/pull/6 . parsebib now gained a set of higher-level parsing functions and the ability to expand strings based on a hash table of `@string` definitions that can be passed to the relevant functions.

The README has the details, but in short: there is now a function `parsebib-parse-buffer` that returns all BibTeX items in a buffer, and there are functions for collecting all entries, `@string` or `@preamble` definitions. Those functions for which it makes sense can take an additional argument to ensure that `@string` abbreviations are expanded. When such an argument is passed, not only are `@string` abbreviations expanded, field values are also returned without enclosing braces or double quotes and sequences of newlines and spaces in field values are reduced to a single space. In essence, passing a hash table of `@string` definitions means that the .bib data is returned in such a way that it becomes suitable for displaying to the user, rather than it being returned exactly how it's present in the .bib file.

The lower-level API can still be used, of course. It now also has the ability to expand `@string` abbrevs and return the data for displaying.

tmalsburg:
This is great, thank you.  I suppose this will allow us to simplify a bunch of things.

Do you think it would also make sense to also add resolution of crossrefs to parsebib?  Conceptually crossrefs are similar to string constants.  By resolution I mean copying fields from the crossreffed entry into the crossreffing entry.  Duplicate fields are allowed but local fields come first in the association list such that they override fields from the crossreffed fields when accessed via `assoc`.

joostkremers:
> This is great, thank you. I suppose this will allow us to simplify a bunch of things.

Yes, instead of using a loop to read all the entries yourself, you could simply call a single function. Note, BTW, that I don't know which kind of format helm expects for its candidates list, so if you'd like me to tweak parsebib's return values in some way, just let me know. I don't have any plans for the near future to use the higher-level API for Ebib, because Ebib does its own error reporting and needs access to the line numbers in the bib file, so I can make parsebib output whatever you like.

> Do you think it would also make sense to also add resolution of crossrefs to parsebib?

Yeah, I guess that would make sense. Ebib already has such a mechanism built in, but it's in a part of the code where it doesn't really make sense. It shouldn't be too difficult to move that code to parsebib.

Would it be important for you that field values that are retrieved from xref'ed entries are marked in some way? If yes, any particular way you'd want it marked? It's important for Ebib, and right now I use a text property on the relevant string.

tmalsburg:
Hey both, I'm currently traveling and don't have time to work on this.  I'll come back to this when I'm back home.  Best wishes!

joostkremers:
I pushed a new version of `parsebib.el` that is able to resolve cross-references. It works slightly differently than resolving `@string` abbrevs and can be used independently from it, although it wouldn't really make sense to do so if only because `@string` abbrevs in the `crossref` field wouldn't be expanded).

Field values that are inherited from cross-referenced fields are not marked in any way (though that would be an easy change). Note that there is support for biblatex's more sophisticated inheritance system as well, though by default, BibTeX's inheritance rule is used.

Details are in the README.

tmalsburg:
Sounds good. In helm-bibtex we don't need marking of inherited fields. They are all treated the same at this time.

malb:
Hiya, since there's some movement in this repo I figured it's okay to send a quick ping on this PR. Anything I should do to make it mergeable?

tmalsburg:
Hi, sorry for not taking care of this.  I hope I will have time to revisit this next week.

tmalsburg:
Hi, I reread the conversations here and over at parsebib, had a look at the code, and conducted some simple tests.  All looks good and I merged.

The code for parsing has become rather comples and I hope we will be able to simplify it in the future using @joostkremers' new high-level functions.

For now, thanks everyone for your contributions!

tmalsburg:
I think there is code for stripping curly braces and quotes around bibtex fields.  Do I understand correctly that that's not needed anymore because parsebib strippes these when provided with a string hash?

tmalsburg:
In 121fb2d720e2a3a3dcdd21eedf3308431318c929 I fixed a small bug that I think is related to this.

joostkremers:
> I think there is code for stripping curly braces and quotes around bibtex fields. Do I understand correctly that that's not needed anymore because parsebib strippes these when provided with a string hash?

Yes, that's correct. Basically, if you pass a string hash to the reading functions, the output is modified so that it becomes suitable for display: braces or double quotes are removed and multiple whitespace characters are reduced to a single space.

PR #162, opened 2016-11-18
allow use of org-mode bibliography (org-bibtex)

ericdanan:
Org-bibtex is a built-in org-mode module that allows the use of org-mode bibliographies: each entry is a heading, and the entry fields and values are stored in its property drawer. Hence org-mode todos, tags, scheduling, etc, can be used seamlessly in the bibliography files. These files can be converted automatically to bibtex files.

With this PR it is now possible to specify org-mode files in `bibtex-completion-bibliography`. In fact very few changes were needed to make this work, because bibliography parsing is still done on the converted bibtex files. The only action that needs to be modified is `bibtex-completion-show-entry`, in order to show the org headline rather than the bibtex entry.

Note that it is also possible to specify an org-mode bibliography file as single notes file, although this is not necessary.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -44,7 +44,11 @@
   :group 'completion)
 
 (defcustom bibtex-completion-bibliography nil
-  "The BibTeX file or list of BibTeX files. Org-bibtex users can also specify org-mode bibliography files, in which case it will be assumed that a BibTeX file exists with the same name and extension bib instead of org. If the bib file has a different name, use a cons cell \"(orgfile.org . bibfile.bib)\" instead."
+  "The BibTeX file or list of BibTeX files. Org-bibtex users can
+also specify org-mode bibliography files, in which case it will
+be assumed that a BibTeX file exists with the same name and
+extension bib instead of org. If the bib file has a different
+name, use a cons cell \"(orgfile.org . bibfile.bib)\" instead."
   :group 'bibtex-completion
   :type '(choice file (repeat file)))
 
@@ -271,21 +275,47 @@ the directories listed in `bibtex-completion-library-path'."
 
 (defcustom bibtex-completion-display-formats
   '((t . "${author:36} ${title:*} ${year:4} ${=has-pdf=:1}${=has-note=:1} ${=type=:7}"))
-  "Alist of format strings for displaying entries in the results list. The key of each element of this list is either a BibTeX entry type (in which case the format string applies to entries of this type only) or t (in which case the format string applies to all other entry types). The value is the format string.
+  "Alist of format strings for displaying entries in the results list.
+The key of each element of this list is either a BibTeX entry
+type (in which case the format string applies to entries of this
+type only) or t (in which case the format string applies to all
+other entry types). The value is the format string.
 
-In the format string, expressions like \"${author:36}\", \"${title:*}\", etc, are expanded to the value of the corresponding field. An expression like \"${author:N}\" is truncated to a width of N characters, whereas an expression like \"${title:*}\" is truncated to the remaining width in the results window. Three special fields are available: \"=type=\" holds the BibTeX entry type, \"=has-pdf=\" holds `bibtex-completion-pdf-symbol' if the entry has a PDF file, and \"=has-notes=\" holds `bibtex-completion-notes-symbol' if the entry has a notes file. The \"author\" field is expanded to either the author names or, if the entry has no author field, the editor names."
+In the format string, expressions like \"${author:36}\",
+\"${title:*}\", etc, are expanded to the value of the
+corresponding field. An expression like \"${author:N}\" is
+truncated to a width of N characters, whereas an expression like
+\"${title:*}\" is truncated to the remaining width in the results
+window. Three special fields are available: \"=type=\" holds the
+BibTeX entry type, \"=has-pdf=\" holds
+`bibtex-completion-pdf-symbol' if the entry has a PDF file, and
+\"=has-notes=\" holds `bibtex-completion-notes-symbol' if the
+entry has a notes file. The \"author\" field is expanded to
+either the author names or, if the entry has no author field, the
+editor names."
   :group 'bibtex-completion
   :type '(alist :key-type symbol :value-type string))
 
 (defvar bibtex-completion-display-formats-internal nil
-  "Stores `bibtex-completion-display-formats' together with the \"used width\" of each format string. This is set internally.")
+  "Stores `bibtex-completion-display-formats' together with the
+\"used width\" of each format string. This is set internally.")
 
 (defvar bibtex-completion-cache nil
-  "A cache storing the hash of the bibliography content and the corresponding list of entries, for each bibliography file, obtained when the bibliography was last parsed. When the current bibliography hash is identical to the cached hash, the cached list of candidates is reused, otherwise the bibliography file is reparsed.")
+  "A cache storing the hash of the bibliography content and the
+corresponding list of entries, for each bibliography file,
+obtained when the bibliography was last parsed. When the
+current bibliography hash is identical to the cached hash, the
+cached list of candidates is reused, otherwise the bibliography
+file is reparsed.")
 
 
 (defun bibtex-completion-normalize-bibliography (&optional type)
-  "Returns a list of bibliography file(s) in `bibtex-completion-bibliography'. If there are org-mode bibliography-files, their corresponding bibtex files are listed as well, unless TYPE is 'main. If TYPE is 'bibtex, org-mode bibliography-files are instead replaced with their associated bibtex files."
+  "Returns a list of bibliography file(s) in
+`bibtex-completion-bibliography'. If there are org-mode
+bibliography-files, their corresponding bibtex files are listed
+as well, unless TYPE is 'main. If TYPE is 'bibtex, org-mode
+bibliography-files are instead replaced with their associated
+bibtex files."
   (delete-dups
    (cl-loop
     for bib-file in (-flatten (list bibtex-completion-bibliography))
@@ -372,7 +402,9 @@ is the entry (only the fields listed above) as an alist."
       append (cddr (assoc file bibtex-completion-cache))))))
 
 (defun bibtex-completion-resolve-crossrefs (files reparsed-files)
-  "Expand all entries with fields from cross-referenced entries in FILES, assuming that only those files in REPARSED-FILES were reparsed whereas the other files in FILES were up-to-date."
+  "Expand all entries with fields from cross-referenced entries
+in FILES, assuming that only those files in REPARSED-FILES were
+reparsed whereas the other files in FILES were up-to-date."
   (cl-loop
    with entry-hash = (bibtex-completion-make-entry-hash files reparsed-files)
    for file in files
@@ -414,7 +446,9 @@ is the entry (only the fields listed above) as an alist."
                   entry)))))
 
 (defun bibtex-completion-make-entry-hash (files reparsed-files)
-  "Return a hash table of all bibliography entries in FILES, assuming that only those files in REPARSED-FILES were reparsed whereas the other files in FILES were up-to-date."
+  "Return a hash table of all bibliography entries in FILES,
+assuming that only those files in REPARSED-FILES were reparsed
+whereas the other files in FILES were up-to-date."
   (cl-loop
    with entries =
      (cl-loop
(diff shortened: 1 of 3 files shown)


bibtex-completion.el
```diff
@@ -1030,16 +1046,20 @@ line."
...
-        (goto-char (point-min))
-        (if (re-search-forward
-             (concat "^@\\(" parsebib--bibtex-identifier
-                     "\\)[[:space:]]*[\(\{][[:space:]]*"
-                     (regexp-quote key) "[[:space:]]*,") nil t)
-            (throw 'break t)
+            (buf (bibtex-completion-buffer-visiting bib-file)))
+        (find-file bib-file)
+        (if (eq major-mode 'org-mode)
+            (when (ignore-errors (org-link-search (concat "#" key)))
```
tmalsburg: You say the the bibliography and the notes file can be the same.  Since we `org-narrow-to-subtree` when showing notes, do we have to `widen` here in order to find the entry?

tmalsburg:
I added two minor comments in the code.  Otherwise it looks very clean.  Thanks for all the amazing contributions.

I discovered org-bibtex only two days ago while looking for packages that are using bibtex.el.  This looks incredibly useful and like a natural way to do bibliographies in Emacs.  I think I'm going to switch. Do you know a tool for converting a bibtex file to org-bibtex?  Should be easy to write one with the infrastructure that we already have in place but I don't want to reinvent the wheel.

bibtex-completion.el
```diff
@@ -284,13 +284,29 @@ In the format string, expressions like \"${author:36}\", \"${title:*}\", etc, ar
   "A cache storing the hash of the bibliography content and the corresponding list of entries, for each bibliography file, obtained when the bibliography was last parsed. When the current bibliography hash is identical to the cached hash, the cached list of candidates is reused, otherwise the bibliography file is reparsed.")
 
 
+(defun bibtex-completion-normalize-bibliography (&optional type)
+  "Returns a list of bibliography file(s) in `bibtex-completion-bibliography'. If there are org-mode bibliography-files, their corresponding bibtex files are listed as well, unless TYPE is 'main. If TYPE is 'bibtex, org-mode bibliography-files are instead replaced with their associated bibtex files."
```
tmalsburg: I think it's standard to wrap doc strings in elisp.

tmalsburg:
Here's a first attempt at a converter bibtex to org-bibtex.  It deviates from org-bibtex' standard format in that it uses `author (year): title` for the headline.  The title alone is not useful in my opinion.  Not sure whether this will lead to conflicts.

``` elisp
(defun bibtex-completion-make-org-bibtex (key)
  (let* ((entry      (bibtex-completion-get-entry key))
         (entry-type (bibtex-completion-get-value "=type=" entry))
         (template 	 "\n* %s (%s): %s\n  :PROPERTIES:\n  :Custom_ID: %s\n  :BTYPE: %s\n%s  :END:\n\n"))
    (format
		 template
		 (if (string= "" (bibtex-completion-apa-get-value "author" entry))
				 (bibtex-completion-apa-get-value "editor" entry)
			 (bibtex-completion-apa-get-value "author" entry))
		 (bibtex-completion-apa-get-value "year" entry)
		 (bibtex-completion-apa-get-value "title" entry)
		 (bibtex-completion-get-value "=key=" entry)
		 entry-type
		 (cl-loop
			for field in entry
			for name = (car field)
			for value = (replace-regexp-in-string "\\(^[\n\t ]*{\\)\\|\\([\n\t ]*}$\\)" "" (cdr field))
			unless (member name '("=type=" "=key=" "=has-pdf=" "=has-note="))
			concat
			(format "  :%s: %s\n" (upcase name) value)))))

(cl-loop
 for entry in (bibtex-completion-candidates)
 for key = (bibtex-completion-get-value "=key=" entry)
 do (insert (bibtex-completion-make-org-bibtex key)))

```

One thing that I don't like about the org-bibtex workflow is that it adds an additional step when importing new entries from the web.  That should be automatized somehow.

bibtex-completion.el
```diff
@@ -284,13 +284,29 @@ In the format string, expressions like \"${author:36}\", \"${title:*}\", etc, ar
...
+  (delete-dups
+   (cl-loop
+    for bib-file in (-flatten (list bibtex-completion-bibliography))
+    for main-file = (if (consp bib-file)
+                        (car bib-file)
+                      bib-file)
+    for bibtex-file = (if (consp bib-file)
+                          (cdr bib-file)
+                        (concat (file-name-sans-extension main-file) ".bib"))
+    unless (equal type 'bibtex)
```
ericdanan: There are three different "normalized" bibliographies we may need:
- only the "main" files (whether .bib or .org), but not the .bib file converted from a .org one. We need this for `bibtex-completion-show-entry`.
- only the bibtex files (so for a .org file, we only want the converted .bib file). We need this for `bibtex-completion-candidates` and `bibtex-completion-get-entry1`.
- all files (so for a .org file we want both this file and its converted .bib file). We need this in `bibtex-completion-init`.

So I chose to assign the types `'main` and `'bibtex` to the first and second case, respectively, and any other value (`nil` in particular) for the third one. Hence line 298 applies whenever it is not `'bibtex`, not only when it is `'main`. Similarly, line 300 applies whenever it is not `'main`, not only when it is `'bibtex`.

But it is surely possible to reformulate this. For instance, we could explicitly assign the type `'full` to the thrid case (and then make the type argument required rather than optional). Then in line 298: `(when (member type '(main full)) ...` and in line 300 `(when (member type '(bibtex full)) ...`. Just let me know how you would like to rephrase things and I will make the change.

bibtex-completion.el
ericdanan: I think you are right as `org-link-search` does not seem to `widen`. I will check and add it if necessary.

ericdanan:
I have to run now but will comment on org-bibtex importing tonight.

bibtex-completion.el
tmalsburg: With point on the doc string `M-x fill-paragraph`.

bibtex-completion.el
tmalsburg: We could ask on the org mailing list.  Perhaps, `org-link-search` actually should widen before searching, at least optionally.  Seems like a common enough use case.

bibtex-completion.el
tmalsburg: I see, I missed the third case where type is nil.  I think we can leave it as it is.

bibtex-completion.el
ericdanan: I confirm that the buffer wasn't widened, so I added `widen`.

Also the function `org-link-search`, when passed a string `"#key"`, basically just calls `(org-find-property "CUSTOM_ID" "key")`. Moreover, org-bibtex has the variable `org-bibtex-key-property`, which defaults to "CUSTOM_ID" but can be set by the user. So I directly used the latter function, as well as the latter variable if bound.

Finally, I realized that `(unless buf (kill-buffer))` at the end of the function was misplaced (it only applied to bibtex files and not to org files).

bibtex-completion.el
ericdanan: Done (also for a bunch of other docstrings I hadn't filled before).

bibtex-completion.el
ericdanan: OK

ericdanan:
AFAIU, org-bibtex's basic mechanism for converting a bibtex entry to an org heading consists of the functions `org-bibtex-read` and `org-bibtex-write`. The former reads the bibtex entry at point, converts it to a field-value alist, and stores this alist in the variable `org-bibtex-entries`. The latter pops the first alist in `org-bibtex-entries`, converts it to an org heading and inserts it at point.

There are a bunch of functions that make use of this mechanism.
- `org-bibtex-read-buffer` and `org-bibtex-read-file` use `org-bibtex-read` to read a whole buffer or file at once and store all entries in `org-bibtex-entries`. 
- `org-bibtex-import-from-file` uses `org-bibtex-read-file` and `org-bibtex-write` to import all entries from a bibtex file as org headings at point.
- `org-bibtex-yank` uses `org-bibtex-read` and `org-bibtex-write` to check if the first item in the kill ring is a bibtex entry and if so insert it at point as an org heading.

So the natural way to convert your bibtex bibliography to org-bibtex is to open an empty org file and call `org-bibtex-import-from-file`. But when I did this my org-bibtex file had a bunch of weird characters. I think that the problem has to do with the fact that my bibtex file is encoded in utf8 and `org-bibtex-read-file` visits the bibtex file literally. In any case, I was able to make it work by visiting my bibtex file "normally" and use `org-bibtex-read-buffer` instead:

```
(defun org-bibtex-import-from-buffer (buffer)
  "Read bibtex entries from BUFFER and insert as Org headlines after point."
  (interactive "bBuffer: ")
  (dotimes (_ (org-bibtex-read-buffer buffer))
    (save-excursion (org-bibtex-write))
    (re-search-forward org-property-end-re)
    (open-line 1) (forward-char 1)))
```

I agree that having the title alone in the headline is not very informative, but this is hardcoded in `org-bibtex-write`. Perhaps the best way to go is to ask the org-bibtex maintainers if they are willing to provide more flexibility? It seems that it would be sufficient to replace the line

```
(insert (funcall val :title))
```

by something like

```
(insert org-bibtex-headline-format)
```

where `org-bibtex-headline-format` would be a variable defaulting to `(funcall val :title)`. Then you could simply get the format you want with:

```
(setq org-bibtex-headline-format (format "%s (%s): %s" (funcall val :author) (funcall val :year) (funcall (val :title)))
```

To import a bibtex entry from the web, it is sufficient to copy it and then call `org-bibtex-yank`, so no additional step w.r.t. standard copy / paste. If we want to insert it at point from the biblio package, then one could add an extended action to biblio that uses `org-bibtex-write` to convert it to an org heading before inserting. Perhaps we could ask the biblio maintainer if he wants to incorporate this action in the package.

To automate generating the .bib file, I simply use the following code:

```
(defvar-local org-bibtex-auto-export nil
  "Whether current file should automatically be exported to a bibtex file with `org-bibtex' each time it is saved.")

(defun org-bibtex-auto-export ()
  "When `org-bibtex-auto-export' is t, export current file to a bibtex file with `org-bibtex', generating the exported file name automatically from the current file name (with extension .bib instead of .org)."
  (when org-bibtex-auto-export
    (org-bibtex (concat (file-name-sans-extension (buffer-file-name))
			".bib"))))

(add-hook 'after-save-hook 'org-bibtex-auto-export)
```

Then if I put

```
# -*- org-bibtex-auto-export: t; -*-
```

at the beginning of an org-bibtex file, the .bib file is generated automatically each time the org file is saved.

tmalsburg:
Thanks for all this information.  Using `org-bibtex-yank` when importing through biblio.el is actually a pretty good solution.  Regarding the format of the title, I think it would be better if the user could specify a function that takes the entry and returns a title string.  Just composing the title from the raw bibtex values is not ideal because these may contain information that I'd like to strip (e.g, curly braces in titles) and I'd also like to shorten the authors to be just a list of surnames.

Your `org-bibtex-auto-export` looks really useful and a variant of this would make a nice addition to org-bibtex.

tmalsburg:
Merged.  I tried converting my bibliography with `org-bibtex-read-buffer` and `org-bibtex-write` but it fails on some entries that look perfectly normal.  Not sure what's going on.  My own code above seems to be more reliable, so I'm gonna use that.  In the process I found that `bibtex-completion-apa-get-value` has a bug and doesn't downcase words in titles when they are outside braces.

ericdanan:
I agree with you on the title format, I didn't write correctly. My idea was that the variable `org-bibtex-headline-format` would hold an arbibrary lisp form (so I guess it should be `(insert (eval org-bibtex-headline-format)` and not `(insert org-bibtex-headline-format)`), but perhaps it would be better to make `org-bibtex-headline-format` a function of one variable (the entry) rather than an arbitrary form.

I'm not sure how to suggest this to the org-bibtex maintainers. It's not on github I'm not sure there's a way to send of pull request. Should we just email the maintainers with the suggestion and code or do you think there's a better way?

I don't know why `org-bibtex-read-buffer` + `org-bibtex-write` fails for you. Your code may not cause problems if some org-bibtex variables are customized, like `org-bibtex-prefix`, `org-bibtex-key-property`, `org-bibtex-tags-are-keywords`, although it seems clearly feasible to make it aware of these variables. Another potential issue is that you won't get the same headline format if you import entries with `org-bibtex-yank`. So being able to modify the format inside `org-bibtex-write` would be the most consistent way in my opinion.

Re `org-bibtex-auto-export`, we could also ask the maintainers if they want to include (or some variant, perhaps allow to specify a filename instead of t) in org-bibtex.

tmalsburg:
> I'm not sure how to suggest this to the org-bibtex maintainers.

I think the standard protocol is to send a patch to the org-mode mailing list.

> Re org-bibtex-auto-export, we could also ask the maintainers if they want to include (or some variant, perhaps allow to specify a filename instead of t) in org-bibtex.

Yes, seems like a feature that many people might want.

ericdanan:
FYI if you update org-mode you can now customize the org-bibtex headline format through the variable `org-bibtex-headline-format-function`.

tmalsburg:
Fantastic, looking forward to using it.

Issue #88, opened 2016-01-21
Multiple PDFs or other file types


ericdanan:
Hello,

Two related suggestions:
1. Handling several PDFs for a given entry. E.g. for a book there could be key-chap1.pdf, key-chap2.pdf... I don't know if this makes sense, but perhaps helm-bibtex-find-pdf-in-library could check for the regexp key*.pdf? And helm-bibtex-open-pdf could start a helm session to choose the file to open? There are other places where the function helm-bibtex-find-pdf-in-library is used, so perhaps this would create complications...
2. Handling other file types than PDF. E.g. DJVU Perhaps using a custom variable to specify which file extensions to search/open?

I am not really fluent in elisp (or any programming language), but if you think this is worth it I can at least try to do something.

In any case, thank you very much for this very useful package.

Eric

tmalsburg:
I like feature requests like that.  Both features are already present.  If you want multiple PDFs, just separate them using semicolons.  Regarding DJVU, helm-bibtex doesn't really care about file types, it just tells Emacs to open the thing whatever it may be.

ericdanan:
I apologize for not being clear. I didn't mention that I don't store the
file path in the bibtex entry but simply have the files in my library path.
Then if I have an entry with key Doe99 in my bibtex file and two files
Doe99-chap1.pdf and Doe-chap2.pdf, helm-bibtex will not spot them.
Similarly if I have a file Doe99.djvu.

If I enter the (absolute) file path in the bibtex entry, then the djvu
problem is solved indeed. But the multiple files issue will remain because
helm-bibtex will always open the first file. My suggestions were about
having helm-bibtex offering to choose which file to open, and have this to
work even if the files are simply in the library path and not in the bibtex
entry (assuming that the file names begin with the bibtex key and have
extension pdf or djvu, say).

Sorry again if I completely misunderstood something.

tmalsburg:
I see.  I don't use multiple PDFs for single entries, so this feature has very low priority for me.  Also there is a risk that it would slow down initial loading of the library, which I would like to avoid.  Some people use helm-bibtex with truely gigantic bibliographies (e.g., crypto.bib) and the recent chances to PDF detection have already slowed down loading quite a bit.  So the bar is high for an implementation of this feature and the utility (in my personal opinion) low.  Perhaps it is easier if you switch to using the `file` field or if you concatenate the chapter PDFs into one book PDF.

[...]

PR #166, opened 2016-12-14
Make 'bibtex-completion-find-pdf-in-field' return only pdf

muahah:
`bibtex-completion-find-pdf-in-field` returned any file in the 'field' entry.
Now it checks the file mime type and only return pdf files.

It is now impossible to have access to other files than pdf, 
maybe a nice helm menu to choose which file you want to open (when they are more than one of them) will be nice ?
I will take a look at that.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -562,6 +562,7 @@ file is specified, or if the specified file does not exist, or if
                           (--map (f-join it path file-name)
                                  (-flatten bibtex-completion-library-path)))) ; Jabref #100
            for result = (-first 'f-exists? paths)
+           if (string-match "application/pdf" (nth 2 record))
            if (not (s-blank-str? result)) collect result)))))))
 
 (defun bibtex-completion-find-pdf-in-library (key-or-entry)


muahah:
Linked to #88,
but finding all files in the bibtex entry should not slow anything down.

tmalsburg:
Why do you want to exclude non-PDF files?  Not everyone is using PDF for articles.

muahah:
I realized that,
but the function is still called `bibtex-completion-find-pdf-in-field`...

I think having a helm menu asking for file to open when multiple files are found can satisfy everyone.
I will open a new pull request for this.
Closing this one.

tmalsburg:
> I realized that,
> but the function is still called bibtex-completion-find-pdf-in-field...

It's called like that for historical reasons.  Not ideal I admit, but some people modify this function and when we rename it their modifications will break.

> I think having a helm menu asking for file to open when multiple files are found can satisfy everyone.
> I will open a new pull request for this.

Isn't a helm menu overkill here?  Helm is for searching in long lists but in this use case we probably will have one file in most cases (the paper) and sometimes two (the paper and a supplement). I think I would prefer a prompt in the minibuffer and no prompt at all when there's only one file.

PR #167, opened 2016-12-18
When multiple PDF files, ask fot the one to open

muahah:
Follow discussion on #166.

When trying to open a file of an entry with multiple associated files, use `completing-read` to ask for the file to open.
@tmalsburg: I cannot find a simpler way than using `completing-read`...

Also add a customizable filter (`bibtex-completion-ignored-files`), to ignore some files based on regular expression (for example ".*\.html" to filter out html pages recorded by default by Zotero).

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -720,13 +720,21 @@ matching PDFs for an entry, the first is opened."
 
 (defun bibtex-completion-open-any (keys)
   "Open the PDFs associated with the marked entries using the
-function specified in `bibtex-completion-pdf-open-function'.  If no PDF is
-found, try to open a URL or DOI in the browser instead."
+function specified in `bibtex-completion-pdf-open-function'.
+If multiple PDFs are found for an entry, ask for the one to
+open using `completion-read'.  If no PDF is found, try to open a URL
+or DOI in the browser instead."
   (dolist (key keys)
-    (let ((pdf (bibtex-completion-find-pdf key)))
-      (if pdf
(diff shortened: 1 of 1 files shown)


tmalsburg:
Thanks for the PR.  I see one problem with this approach: Lets assume that you configured `bibtex-completion-ignored-files` to ignore HTML files.  Now, when an entry has only an HTML file attached, helm-bibtex will show the symbol indicating that a document is present but when the user tries to open it, it won't because it is ignored.  If we want this feature (I'm not entirely convinced yet, see below), we'd need a solution that goes deeper.

You mentioned that Zotero attaches HTML files that you apparently don't ever want to open in helm-bitex.  I think in this case the problem really lies with Zotero which shouldn't attach these files in the first place.  Adding a workaround in helm-bibtex is possible but may not be the preferred solution.  So the question is: Why does Zotero attach these files? And is it possible to prevent this from happening?  What's the content and purpose of these files anyway?

muahah:
Well,

Zotero offers the possibility to record entries from web pages.
if the web page is related to a an article, Zotero search for DOI, make the associated entry and record the web page (HTML or others) as an associated file (so you can visit it back i guess...).
After verification, it can be deactivated.

So I think you are right, better have clean entries in Zotero than filters in helm-bibtex.
I removed the filter in the new version, I still think asking which file to open when more than more are present can be useful (for the 'one pdf per chapter' situation in #88 for example).

tmalsburg:
Merged.  Thanks.

PR #171, opened 2017-01-03
Fix bibtex-completion-candidates order

justbur:
With nreverse the default ordering in ivy and helm is reverse alphabetical.


tmalsburg:
There is no sorting, alphabetical or otherwise.  The displayed order is the reverse of the order of entries in the Bibtex file.  So if you add a new entry at the bottom of the file, it shows up at the top of the results.  It's a feature not a bug :-)

justbur:
I always thought it was odd that the entries showed up in reverse alphabetical order. I guess the tools I use to add entries keep them sorted. 

I'll close this, but for aesthetic purposes I think the option to sort them would be nice.

tmalsburg:
Sorting has been on my wish list for a long time but I never got around implementing it.

PR #177, opened 2017-02-22
defcustom uses wrong type for some variables

plantarum:
Customization variables `bibtex-completion-addtional-search-fields` and `bibtex-completion-no-export-fields` are defined as `:type 'list`. However, they should be `:type '(repeat string)`. The `list` type is intended for use with a list of fixed length, where each element is a set type (and potentially a different type). In contrast, `repeat` type is for lists of variable length, but each element is the same type.

The interface as is does work, but the customization interface is more helpful with the correct type. As it is now:

```
Hide bibtex-completion-additional-search-fields: nil
   [ State ]: STANDARD. (mismatch)
   The fields that are used for searching in addition to author, Hide
   editor, title, year, BibTeX key, and entry type.
Groups: [Bibtex Completion]
```

Using the repeat definition:
```
Hide Bibtex Completion Additional Search Fields:
[INS]
   [ State ]: STANDARD.
   The fields that are used for searching in addition to author, Hide
   editor, title, year, BibTeX key, and entry type.
Groups: [Bibtex Completion]
```

There are two key differences - Emacs reports a mismatch in the first version, and the second version provides an [INS] widget to add new strings to the list.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -218,13 +218,13 @@ browser in `helm-browse-url-default-browser-alist'"
   "The fields that are used for searching in addition to author,
 editor, title, year, BibTeX key, and entry type."
   :group 'bibtex-completion
-  :type 'list)
+  :type '(repeat string))
 
 (defcustom bibtex-completion-no-export-fields nil
   "A list of fields that should be ignored when exporting BibTeX
 entries."
   :group 'bibtex-completion
-  :type 'list)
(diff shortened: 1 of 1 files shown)


plantarum:
Found another mismatch in bibtex-completion-notes-path. The default value of nil was not included in the list of choices, which somehow confused the widget.

tmalsburg:
Merged, thank you.

PR #183, opened 2017-03-08
Escape regexp chars to avoid creating malformed/mismatched regexps

ebpa:
Escape/quote special regexp characters in bibtex keys to avoid inadvertently mis-translating the keys when creating this search regexp.

Typically this is not encountered since keys ideally omit punctuation, but a wild "[" in a key yields an invalid regexp or a "*" likely yields a non-matching regexp.  In practicality, this only likely to occur with dirty data, but this change improves the durability of the code.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -617,7 +617,7 @@ find a PDF file."
                                 (save-restriction
                                   (widen)
                                   (goto-char (point-min))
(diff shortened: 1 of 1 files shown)


tmalsburg:
Thank you. I found another instance of this: 2e83867a39bbd4b7cb7901b897a66a131b5635e6

ebpa:
Great!  Thanks!  :+1:

PR #185, opened 2017-03-19
extract URL from note field if present

malb:
[crypto.bib](https://cryptobib.di.ens.fr) likes URLs in note fields, e.g.:

```
@Misc{EPRINT:Nam08,
  author =       "Junghyun Nam",
  title =        "Infringing and Improving Password Security of a Three-Party Key Exchange Protocol",
  year =         2008,
  howpublished = "Cryptology ePrint Archive, Report 2008/065",
  note =         "\url{http://eprint.iacr.org/2008/065}",
}
```

This PR adds support for it.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -702,11 +702,23 @@ matching PDFs for an entry, the first is opened."
       (-each it bibtex-completion-pdf-open-function)
     (message "No PDF(s) found.")))
 
+(defun bibtex-completion-extract-url (field entry)
+  "Extract URL from FIELD of ENTRY.
+
+Return nil if no URL is found."
+  (let* ((note (bibtex-completion-get-value field entry)))
+    (if note
+        (when (string-match "https?://" note)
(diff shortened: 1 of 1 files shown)


tmalsburg:
The canonical way to do this is to store the urls in the url field:

```
@Misc{EPRINT:Nam08,
  author =       "Junghyun Nam",
  title =        "Infringing and Improving Password Security of a Three-Party Key Exchange Protocol",
  year =         2008,
  howpublished = "Cryptology ePrint Archive, Report 2008/065",
  url =          "http://eprint.iacr.org/2008/065",
}
```

I'm not in favor of supporting the peculiarities of individual data bases, mainly because there are so many and we'd be opening a can of worms.  I think it would be better to file an issue with crypto-bib and to convince them to change their format.  A simple search-and-replace may be enough.

malb:
Fair enough.

tmalsburg:
Thanks for your understanding.  If copy & replace doesn't cut it, parsebib's new parsing facilities might help with rewriting the entries.

Issue #186, opened 2017-03-20
Insert Reference with long author lists produces strange output


jschwab:
This uses the same bibtex entry as in https://github.com/tmalsburg/helm-bibtex/pull/163 .

At present, the Insert Reference action produces

    - Paxton, B., Marchant, P., Schwab, J., Bauer, E.~B., Bildsten, L.,
      Cantiello, M., Dessart, L., Farmer, R., Hu, H., Langer, N.,
      Townsend, R.~H.~D., Townsley, D.~M., Timmes, F.~X.,
      .. (2015). Modules for Experiments in Stellar Astrophysics (MESA):
      Binaries, Pulsations, and Explosions. \apjs, 220(),
      15. http://dx.doi.org/10.1088/0067-0049/220/1/15

I just noticed the strange trailing `..` before the year.

The function `bibtex-completion-apa-format-authors` seems like it wants to handle long author lists.  When nauthors > 8, it appends `', ...'`, but it still displays all of the authors.  (It just doesn't include the ampersand.)

https://github.com/tmalsburg/helm-bibtex/blob/master/bibtex-completion.el#L966

Then, this gets mangled later on by a punctuation-replacing regexp

https://github.com/tmalsburg/helm-bibtex/blob/master/bibtex-completion.el#L902

which leads to the weird `..` output.

tmalsburg:
Thanks for reporting. Should be fixed.  Now using the Unicode ellipsis glyph instead of `...`.

PR #196, opened 2017-07-28
formatting references truncated the first author

plantarum:
In the function bibtex-completion-apa-format-authors, long author lists
were truncated to eight authors using the function -slice. However, the
call to slice used the hard-coded arguments 1 and 8 to delimit the first
and last authors to retain. This incorrect, as the first list element
is 0 in -slice (i.e., it is 0-indexed.)


plantarum:
The problem was introduced in #186

tmalsburg:
Ah, thanks.  I think this was interference from R where indexing starts at 1 instead of 0.

Issue #198, opened 2017-08-08
bibtex-completion doesn't work in Emacs 26


chrikoehn:
Somehow the alias `acons` (for `cl-acons`) isn't defined in Emacs 26 (I'm using 26.0.50) anymore. I don't know whether this is a bug or a feature in Emacs itself. Anyway, the alias was originially defined for backwards-compatibility with `cl.el` AFAIK and therefore I suggest to replace it with `cl-acons`.

tmalsburg:
Merged.

PR #199, opened 2017-08-08
remove alias acons, replace with cl-acons #198 

chrikoehn:
Issue #198


tmalsburg:
Awesome.  Thanks for the PR.

PR #204, opened 2017-09-22
find local bibliography more robustly

### LINKED ISSUE #184: bibtex-completion-find-local-bibliography should also search in master file
bibtex-completion-find-local-bibliography searches in the directory of the master file for files referenced by \bibliography commands found in the current buffer.
However, if there are multiple *.tex files, \bibliography is often found in the master file. Thus the master file should be searched as well. I'm currently using this hack:

```
(defun bibtex-completion-find-local-bibliography ()
  "Return a list of BibTeX files associated with the current
file. If the current file is a BibTeX file, return this
file. Otherwise, try to use `reftex' to find the associated
BibTeX files. If this fails, return
`bibtex-completion-bibliography'."
  (or (and (buffer-file-name)
           (string= (or (f-ext (buffer-file-name)) "") "bib")
           (list (buffer-file-name)))
      (and (buffer-file-name)
           (require 'reftex-parse nil t)
           (let ((file (file-name-nondirectory (if (TeX-local-master-p)
                                                   (if (f-ext\? TeX-master)
                                                       TeX-master
                                                     (concat TeX-master ".tex"))
                                                 (buffer-name)))
                       )
                 (buf (current-buffer)))
             (unless (get-buffer file)
               (progn
                 (find-file (concat (TeX-master-directory) "/" file))
                 (switch-to-buffer buf)))
             (with-current-buffer file
               (reftex-locate-bibliography-files
                (if (fboundp 'TeX-master-directory)
                    (TeX-master-directory)
                  (file-name-directory (buffer-file-name))))
               )
             )
           )
      bibtex-completion-bibliography))
```

ericdanan:
Using function `reftex-get-bibfile-list` from reftex-cite.el instead
of `reftex-locate-bibliography-files` from reftex-parse.el. This is
more robust because bibliography commands are found in all the
document's files (master file and others), not only in the current buffer.

I think this should fix #184.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -1191,11 +1191,8 @@ BibTeX files. If this fails, return
            (string= (or (f-ext (buffer-file-name)) "") "bib")
            (list (buffer-file-name)))
       (and (buffer-file-name)
-           (require 'reftex-parse nil t)
-           (reftex-locate-bibliography-files
-            (if (fboundp 'TeX-master-directory)
-                (TeX-master-directory)
-              (file-name-directory (buffer-file-name)))))
+           (require 'reftex-cite nil t)
+           (reftex-get-bibfile-list))
       bibtex-completion-bibliography))
 
 (provide 'bibtex-completion)


tmalsburg:
Merged.  Thank you very much.

Issue #121, opened 2016-08-05
Open PDF for bib entry containing point (enhancement request)


higham:
It would be very nice if there was a command to open the PDF for the bib entry the point  is located within (i.e., in the bib file).  At the moment the only way I can see to do this is to move the point to the key, call helm-bibtex, press M-n, then hit enter.

tmalsburg:
This sounds useful indeed.  Unfortunately, I won't have time to work on this over the next couple of weeks or even months (new job, transatlantic move, ...).  It shouldn't be too hard, though, to implement this.  bibtex.el probably has tools in place for extracting the bibtex key from the entry at point and once you have that you can pass it to `bibtex-completion-open-pdf`.  PR welcome.

jagrg:
I'm not sure if bibtex.el has anything specific to extract the key. Maybe I didn't look close enough, but here's a starting point:

```elisp
(defun bibtex-open-pdf ()
  (interactive)
  (save-excursion
    (bibtex-beginning-of-entry)
    (let* ((key (reftex-get-bib-field "=key=" (bibtex-parse-entry t)))
	   (pdf (car (bibtex-completion-find-pdf-in-library key))))
      (if pdf
	  (funcall bibtex-completion-pdf-open-function pdf)
	(message "No PDF found for %s" key)))))
```

tmalsburg:
Great, thanks @jagrg.  Here's a version that does the same but without using reftex:

``` elisp
(defun bibtex-completion-open-pdf-of-entry-at-point ()
  (interactive)
  (save-excursion
    (bibtex-beginning-of-entry)
    (when (looking-at bibtex-entry-maybe-empty-head)
      (bibtex-completion-open-pdf (bibtex-key-in-head)))))
```

@higham, does this work for you?

higham:
Yes - bibtex-completion-open-pdf-of-entry-at-point works perfectly!
Thanks to both of you.  This is something I will use multiple times per day.

tmalsburg:
Glad to hear.  Now the question is how and whether we should integrate this into helm-bibtex since this is slightly outside its scope.  For example, should we assign a default key binding?

higham:
Personally, I would say yes. I've set C-c p. I've also made a trivial
modification of the macro to call bibtex-completion-edit-notes on the
current bib entry, assigning it to C-c n. Of course, I now also want to be
able to do both actions with the point on the cite command in the LaTeX
file!  The obvious, rather crude code

[...]

PR #205, opened 2017-09-27
preselect bibtex entry at point

ericdanan:
Re #121 

The core function is `bibtex-completion-key-at-point`, which identifies the bibtex key around point depending on the context. Here is its docstring:

> "Return the key of the BibTeX entry at point. If the current
file is a BibTeX file, return the key of the entry at
point. Otherwise, try to use reftex to check whether point is
at a citation macro, and if so return the key at
point. Otherwise, if the current file is an org-mode file, return
the value of `org-bibtex-key-property' (or
default to \"CUSTOM_ID\"). Otherwise, return nil."

It can certainly be enriched, eg to detect if point is on a crossref field or an org-mode citation link.

Then `helm-bibtex` and `ivy-bibtex` make use of it to preselect the corresponding entry. For `helm-bibtex` this sometimes implies increasing the candidate number limit.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -1195,6 +1195,38 @@ BibTeX files. If this fails, return
            (reftex-get-bibfile-list))
       bibtex-completion-bibliography))
 
+(defun bibtex-completion-key-at-point ()
+  "Return the key of the BibTeX entry at point. If the current
+file is a BibTeX file, return the key of the entry at
+point. Otherwise, try to use `reftex' to check whether point is
+at a citation macro, and if so return the key at
+point. Otherwise, if the current file is an org-mode file, return
+the value of `org-bibtex-key-property' (or
+default to \"CUSTOM_ID\"). Otherwise, return nil."
+  (or (and (eq major-mode 'bibtex-mode)
+           (save-excursion
+             (bibtex-beginning-of-entry)
+             (and (looking-at bibtex-entry-maybe-empty-head)
+                  (bibtex-key-in-head))))
+      (and (require 'reftex-parse nil t)
+           (save-excursion
+             (skip-chars-backward "[:space:],;}")
+             (let ((macro (reftex-what-macro 1)))
+               (and (stringp (car macro))
+                    (string-match "\\`\\\\cite\\|cite\\'" (car macro))
+                    (thing-at-point 'symbol)))))
+      (and (eq major-mode 'org-mode)
+           (let (key)
+             (and (setq key (org-entry-get nil
+                                           (if (boundp 'org-bibtex-key-property)
+                                               org-bibtex-key-property
+                                             "CUSTOM_ID")
+                                           t))
+                  ;; KEY may be the empty string the the property is
+                  ;; present but has no value
+                  (> (length key) 0)
+                  key)))))
+
 (provide 'bibtex-completion)
 
 ;; Local Variables:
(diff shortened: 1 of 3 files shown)


tmalsburg:
Looks good but I added a comment (with this review thing).

ericdanan:
Sorry to ask but where can I find your review? I just see "No reviews" in the conversation page...

tmalsburg:
Weird, I see the review thing above this discussion.  Anyway, in `helm-source-bibtex` you are referencing `helm-bibtex-candidates` which (as far as I can see) is not defined.  The current code uses `bibtex-completion-candidates`.

ericdanan:
This is to avoid calling `bibtex-completion-candidates` twice, because now we also need the candidates to compute the index of the entry to be preselected (perhaps it is possible to do this inside the source, but I didn't find a way to do so). So my strategy is to let-bind `candidates` to the list returned by this function, use it to compute the preselect index, and finally call `helm` and make it pass `candidates` to the source. This last step is done by the `:bibtex-candidates` keywords which locally binds the variable `helm-bibtex-candidates` inside the helm session.

jagrg:
It looks fine so far in terms of speed, but it fails to find candidates that have cross-references in the buffer. In my example, I used  `(when (re-search-forward (concat bibtex-entry-maybe-empty-head (regexp-quote key))` ... to skip cross-references when searching, but maybe there's a better solution.

ericdanan:
@jagrg do you mean if you are in a bibtex file with point on an entry with key `XXX` and a crossref field with value `YYY`, then the `XXX` entry is not preselected? Is `YYY` preselected then or no entry is preselected?

jagrg:
I mean if you are in a bibtex file with point on an entry which is cross-referenced somewhere above that entry. What happens is it selects the first candidate referencing the entry at point.

jagrg:
Actually it only fails if the entry at point is the last one in the buffer (and cross referenced right above).

ericdanan:
Thanks, should be fixed now.

It was not specifically related to cross-references. The general issue was that when the candidate at point was the first one in the candidates list, `helm-bibtex` would preselect the second candidate instead. `bibtex-completion-key-at-point` would find the correct entry and `cl-position-if` would correctly find its index 0 in the list, but `(helm-preselect 0)` would move one line forward (the number passed as argument is discarded unless it is > 0). `ivy` has a different preselection mechanism (can directlly preselect a candidate by its index) and was not affected.

jagrg:
You're right. Thanks for the explanation. It works with 3c83073.

On 29 Sep 2017, Eric Danan <notifications@github.com> wrote:

> Thanks, should be fixed now.
>
> It was not specifically related to cross-references. The general issue was that when the candidate at point was the first one in the candidates list, `helm-bibtex` would preselect the second candidate instead. `bibtex-completion-key-at-point` would find the correct entry and `cl-position-if` would correctly find its index 0 in the list, but `(helm-preselect 0)` would move one line forward (the number passed as argument is discarded unless it is > 0). `ivy` has a different preselection mechanism (can directlly preselect a candidate by its index) and was not affected.


-- 
Jonathan
Monthly Atmospheric CO₂: August 2017: 405.07, August 2016: 402.25

tmalsburg:
> This last step is done by the :bibtex-candidates keywords which locally binds the variable helm-bibtex-candidates inside the helm session.

Thanks for clarifying, I wasn't aware of this mechanism.  All looks good then and I'll merge this PR.  Thanks very much to both Eric and Jonathan!

PR #206, opened 2017-09-27
ignore reftex errors when finding local bibliography

ericdanan:
When `bibtex-completion-find-bibliography` resorts to `reftex-get-bibfile-list` if it does not find a bibliography file. Ignoring such errors allows to fall back to the global bibliography.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -1192,7 +1192,7 @@ BibTeX files. If this fails, return
            (list (buffer-file-name)))
       (and (buffer-file-name)
            (require 'reftex-cite nil t)
-           (reftex-get-bibfile-list))
+           (ignore-errors (reftex-get-bibfile-list)))
       bibtex-completion-bibliography))
 
 (provide 'bibtex-completion)


tmalsburg:
This makes sense given the current implementation, but I do wonder whether the current implementation is sensible.  Currently, `helm-bibtex-with-local-bibliography` falls back to the global bibliography when no local bibliography can be found.  However, when this function is used the user probably really wants something else than the global bibliography.  Perhaps, it's better to fail with an error when a local bibliography can't be found? Scenario: User wants to check bibliography associated with current manuscript but what they actually get to see is their global bibliography.

ericdanan:
I personally find it convenient to have a single keybinding to open the local bibliography if there is one and the global one otherwise, but I understand your scenario as well.

Perhaps then a more flexible approach would be to no longer fall back to the global bibliography by default but add an optional argument `fallback-to-global` to `helm/ivy-bibtex-with-local-bibliography`. What do you think?

tmalsburg:
You're right that falling back to global is useful, it's just not what the name of the function suggests.  But perhaps this can be solved by using a headline for the source (helm has that, not sure about ivy) that indicates what data we're looking at.

ericdanan:
Yes we could append "(local/global bibliography)" to the `helm-bibtex` source name. For `ivy-bibtex` we could put this in the minibuffer prompt.

I can do that, but I'd rather wait until you decide what to do with #205 to avoid merge conflicts.

PR #207, opened 2017-09-29
properly fallback to global bibliography

ericdanan:
Replaces #206

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -1185,15 +1185,13 @@ editing."
   "Return a list of BibTeX files associated with the current
 file. If the current file is a BibTeX file, return this
 file. Otherwise, try to use `reftex' to find the associated
-BibTeX files. If this fails, return
-`bibtex-completion-bibliography'."
+BibTeX files. If this fails, return nil."
   (or (and (buffer-file-name)
            (string= (or (f-ext (buffer-file-name)) "") "bib")
            (list (buffer-file-name)))
       (and (buffer-file-name)
            (require 'reftex-cite nil t)
-           (reftex-get-bibfile-list))
-      bibtex-completion-bibliography))
+           (ignore-errors (reftex-get-bibfile-list)))))
 
 (defun bibtex-completion-key-at-point ()
   "Return the key of the BibTeX entry at point. If the current
--- a/helm-bibtex.el
+++ b/helm-bibtex.el
@@ -179,6 +179,8 @@ comes out in the right buffer."
 
 (defvar helm-source-bibtex
   (helm-build-sync-source "BibTeX entries"
+    :header-name (lambda (name)
+                   (format "%s%s: " name (if helm-bibtex-local-bib " (local)" "")))
     :candidates 'helm-bibtex-candidates
     :filtered-candidate-transformer 'helm-bibtex-candidates-formatter
     :action (helm-make-actions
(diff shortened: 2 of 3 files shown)


tmalsburg:
This is great! I think I will now make the local version of helm-bibtex my default.

At an earlier occasion you added support for org-bibtex.  I suppose we should at some point also support org-bibtex in `bibtex-completion-find-local-bibliography`, i.e. show the content of the local file when we're in an org-bibtex file.  Not sure though how we can reliably detect an org-bibtex file.  (I'm not really familiar with org-bibtex.)

ericdanan:
Perhaps by checking that the function `org-bibtex` is bound and calling it. It will export all bibtex entries it finds to a bibtex file. If this file is empty we can consider that there is no local bibliography. If it is not we can parse it to get the candidates.

But I am not sure it is a good idea to write files to disk like this, because it uses space and memory. And it could be redundant if the user is already exporting the org-bibtex file to a bibtex file (perhaps to use it for a tex document), but we can't know from the org file where to find this bib file.

It may be possible to avoid these issues by writing a function to parse org files directly (which could also be used for the global bibliography by the way). I don't know if there would be a speed issue wrt parsing a bib file for large files.

tmalsburg:
I agree that we shouldn't write files.  Re parsing org directly: If there is code to produce BibTeX such a parser must already exist, no?

Another thing: Currently we're showing either the local or the global bib, but wouldn't it be better to always show the global bib and the local one in addition if present?  In Helm this makes total sense, since we can have multiple sources separated by headlines.

ericdanan:
> Re parsing org directly: If there is code to produce BibTeX such a parser must already exist, no?

org-bibtex has the function `org-bibtex-headline`. It basically parses the bibtex properties and tags for the org-mode entry at point into an alist and returns the corresponding bibtex entry as a string. Then the function `org-bibtex` calls `org-bibtex-headline` for each org-mode entry, concatenates the bibtex entry strings, and write the concatenated string to a bibtex file.

So a possibility could be to write a function like `org-bibtex` but simply returning the concatenated string without writing any file. Then we could write this string into a temp buffer and parse it with `bibtex-completion-parse-bibliography`. But it seems inefficient to parse the same information twice: first from the org-mode entry and second from the corresponding bibtex entry. We would need to submit a patch to org-bibtex to break down `org-bibtex-headline` into two functions so we can directly use the list parsed from the org-mode entry. Then another issue would be that in org-bibtex the alist keys are not strings but keywords (`:author` `:=key=` ...), but this is perhaps not a big deal.

> Currently we're showing either the local or the global bib, but wouldn't it be better to always show the global bib and the local one in addition if present? In Helm this makes total sense, since we can have multiple sources separated by headlines.

It seems feasible, but perhaps it would be too slow for a user who has a large global bibliography and just wants the local one? Also we would have to be careful when an action calls `bibtex-completion-get-entry`, we neeed to keep track in which bibliography to look.

tmalsburg:
Re org-bibtex: one could probably use `org-bibtex-headline` as a template for a function that creates an entry in bibtex-completion format but it's not a priority for me since I'm not (yet) using org-bibtex.

Re showing local and global: I think the slowdown should not be a problem, since the global bibliography will be in the cache already (in the typical case anyway).  But you are right, if a key is used in the local and the global bib, we don't know what to do.

PR #210, opened 2017-10-02
allow using date field instead of year

ericdanan:
For biblatex the date field is preferred to year. So when parsing the bibliography, record either date or year, and when formatting candidates for display, extract the year from the date field if therer is no year field. Similarly extract the year from date if needed in `bibtex-completion-format-citation-org-apa-link-to-PDF`.

I didn't touch `bibtex-completion-apa-get-value`, I guess it would make sense to do the same there but I noticed that you are using editor as an alternative to author there so I'm not sure what you would like to do.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -942,6 +942,8 @@ guidelines.  Return DEFAULT if FIELD is not present in ENTRY."
          ;; the journal in its title.
          ("pages" (s-join "–" (s-split "[^0-9]+" value t)))
          ("doi" (s-concat " http://dx.doi.org/" value))
+         ("year" (or value
+                     (car (split-string (bibtex-completion-get-value "date" entry) "-"))))
          (_ value))
       "")))
 
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -482,17 +482,21 @@ return a list of entries in the order in which they appeared in
 the BibTeX file. Also do some preprocessing of the entries."
   (goto-char (point-min))
   (cl-loop
-   with fields = (append '("title" "year" "crossref")
+   with fields = (append '("title" "crossref")
                          (-map (lambda (it) (if (symbolp it) (symbol-name it) it))
                                bibtex-completion-additional-search-fields))
    for entry-type = (parsebib-find-next-item)
    while entry-type
    unless (member-ignore-case entry-type '("preamble" "string" "comment"))
    collect (let* ((entry (parsebib-read-entry entry-type))
-                  (fields (cons (if (assoc-string "author" entry 'case-fold)
-                                    "author"
-                                  "editor")
-                                fields)))
+                  (fields (append
+                           (list (if (assoc-string "author" entry 'case-fold)
+                                     "author"
+                                   "editor")
+                                 (if (assoc-string "date" entry 'case-fold)
+                                     "date"
+                                   "year"))
+                           fields)))
              (-map (lambda (it)
                      (cons (downcase (car it)) (cdr it)))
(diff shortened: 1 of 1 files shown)


tmalsburg:
Not sure I understand your last point but I think that it makes sense for `bibtex-completion-apa-get-value` to return information from the year field or the date field when asked for "year".  That's also what you're proposing, right?

ericdanan:
Sorry my last point was not clear. I meant that when looking at the functions `bibtex-completion-apa-format-reference` and `bibtex-completion-edit-notes`, which both rely on `bibtex-completion-apa-get-value`, it seems to me that `author` there is not replaced with `editor` if not present. So I don't know whether this should be fixed or if there is a reason to be strict with this field (and so possibly also for `year`) there.

If we want to use date as an alternative to year there, then a possibility is to modify `bibtex-completion-apa-get-value` to do the job. But now I'm wondering if it would perhaps be simpler and more powerful to modify `bibtex-completion-get-value` to accept strings such as `author|editor` or `date|year`, ie use | as an "or" operator and try fields in sequence until one returns a value or the last one is reached.

tmalsburg:
The reason for being strict with authors and editors is that they can't be used interchangeably in APA references. For instance, when you cite a proceedings paper in APA style, you list the authors and the editors, so different semantics and you wouldn't insert the editors in the position of the author if authors were missing.  For year, the difference is not semantics but just were the information is coming from.  For that reason I believe that year could be handled differently in `bibtex-completion-apa-get-value`.  Does this make sense?

Interesting idea re making `bibtex-completion-get-value` more flexible. Though I'm not sure how much this would buy us.  There are few places where this is needed and you'd still have to do extra work in the case of `year|date` because you'd have to extract the year from the date. Unless this simplifies other things as well, I wouldn't add that capability.

ericdanan:
> The reason for being strict with authors and editors is that they can't be used interchangeably in APA references. For instance, when you cite a proceedings paper in APA style, you list the authors and the editors, so different semantics and you wouldn't insert the editors in the position of the author if authors were missing. For year, the difference is not semantics but just were the information is coming from. For that reason I believe that year could be handled differently in bibtex-completion-apa-get-value. Does this make sense?

It makes sense. But what happens then if you use `bibtex-completion-apa-format-reference` (eg called by `bibtex-completion-edit-notes`) for a book that has editors but no author?

> Interesting idea re making bibtex-completion-get-value more flexible. Though I'm not sure how much this would buy us. There are few places where this is needed and you'd still have to do extra work in the case of year|date because you'd have to extract the year from the date. Unless this simplifies other things as well, I wouldn't add that capability.

I agree it won't buy much, so will simply add a couple of lines to `bibtex-completion-apa-get-value` to handle the `year` field.

tmalsburg:
> It makes sense. But what happens then if you use bibtex-completion-apa-format-reference (eg called by bibtex-completion-edit-notes) for a book that has editors but no author?

Good point. This is something that needs to be improved at some point in the future.  My implementation of the APA standard (which is horribly complex) is really just cartoon version of the real thing.

> I agree it won't buy much, so will simply add a couple of lines to bibtex-completion-apa-get-value to handle the year field.

Great.

ericdanan:
Done.

BTW there seems to be something wrong in `bibtex-completion-apa-get-value` for the `editor` field (line 921 onwards):

```
         ("editor"
          (if (string= entry-type "proceedings")
              (bibtex-completion-apa-format-editors value)
            (bibtex-completion-apa-format-editors value)))
```

tmalsburg:
Thanks for amending the PR.  And yes, I also spotted the thing with the editor field. Not sure what I was intending to do there.

PR #211, opened 2017-10-02
make cross-referenced entry types customizable

ericdanan:
Maybe I'm the only one interested in this...

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -296,14 +296,13 @@ editor names."
   :group 'bibtex-completion
   :type '(alist :key-type symbol :value-type string))
 
-(defcustom bibtex-completion-cross-referenced-entry-types '("proceedings" "book")
+(defvar bibtex-completion-cross-referenced-entry-types
+  '("proceedings" "mvproceedings" "book" "mvbook" "collection" "mvcollection")
   "The list of potentially cross-referenced entry types (in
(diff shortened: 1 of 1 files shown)


tmalsburg:
Out of curiosity what is your setting going to be for this variable?

ericdanan:
First I would add `collection` and the biblatex types `mvbook`, `mvcollection` and `mvproceedings`.

Second I would add `article`. I don't think articles are really cross-referenced in practice, but I happen to use a bibtex file to keep track of reports that I make, and so cross-referencing would be useful to handle revisions (of course I don't cite anything from this file, but I find it to be a convenient way to store the papers and reports or search for a paper I've reviewed in the past). As I said, this may not interest anyone except me.

tmalsburg:
I think it would make sense to add collection, mvcollection, mvbook, mvproceedings to the entry types that are stored in the hash by default.  I just didn't add them because I don't use them and didn't even know some of them. Second, perhaps we could introduce this new variable as a `defvar` instead of a `defcustom` because most people will never change it (I assume). Third, `bibtex-completion-crossreferenced-entry-types` should probably be mentioned in the doc string of `bibtex-completion-make-entry-hash`. Otherwise we can merge this.

ericdanan:
Done.

tmalsburg:
Great, thank you!

PR #213, opened 2017-10-13
Handle multiple PDFs or other file types

ericdanan:
See #88.

Added a custom variable `bibtex-completion-find-additional-pdfs`, defaulting to nil so nothing is changed by default. When it is non-nil, all pdf files whose basename starts with the bibtex key (e.g. `<key>-annotated.pdf` or `<key>-supplemental.pdf`) are considered as PDFs. But this only happens in the action functions. There is no performance impact when building the bibliography because only `<key>.pdf` is considered then (the assumption is that additional files only exist when that file does). In my testing, looking for all PDFs for an entry in a directory of 10,000 files took 0.15 seconds, which seems acceptable.

When opening an entry's PDF, the user is prompted for the file to select if there are more than one (to this end I moved some code from `bibtex-completion-open-any` to `bibtex-completion-open-pdf` and made the former rely on the latter). when attaching an entry's PDF, all PDFs are attached. When adding a entry's PDF to the library, the user is prompted for the filename with default `<key>.pdf`.

Also created a defcustom `bibtex-completion-pdf-extension`, defaulting to `".pdf"`, which allows to change the file type. Note that setting it to `"-annotated.pdf"` or `"-supplemental.pdf"` is a simple way to handle annotated PDFs or supplemental material as discussed in the `README`, but I guess this is less powerful than setting `bibtex-completion-find-additional-pdfs`. More interestingly, it is also possible to give a list of extensions (e.g. `'(".pdf" ".djvu")`), which are then tried sequentially until a file is found. This does slow down building the library a bit, since adding a second filetype induces an extra call to `f-file?` for all entries that don't have a PDF. So someone with a large bibliography and few pdf files may not want to do that. In my testing, calling `f-file?` 10,000 times took `0.8` second, and my bibliography is not that large so it is acceptable for me.

I didn't update the `README` but I can do so if you are interested in this PR.

### DIFF
--- a/README.org
+++ b/README.org
@@ -7,6 +7,7 @@ Ivy-bibtex: [[http://melpa.org/#/ivy-bibtex][http://melpa.org/packages/ivy-bibte
 Helm-bibtex and ivy-bibtex allow you to search and manage your BibTeX bibliography.  They both share the same generic backend, bibtex-completion, but one uses the Helm completion framework and the other Ivy as a front-end.
 
 * News
+- 2017-10-21: Added support for multiple PDFs or other file types.
 - 2017-10-02: Use date field if year is not defined.
 - 2017-09-29: If there is a BibTeX entry, citation macro, or org-bibtex entry at point, the corresponding publication will be pre-selected in helm-bibtex and ivy-bibtex giving quick access to PDFs and other functions.
 - 2016-11-24: Added support for bare relative paths to PDF files.  Concatenates the path in the ~file~ field to all paths in ~bibtex-completion-library-path~.
@@ -90,7 +91,7 @@ Specify where PDFs can be found:
 (setq bibtex-completion-library-path '("/path1/to/pdfs" "/path2/to/pdfs"))
 #+END_SRC
 
-Bibtex-completion assumes that the name of a PDF consists of the BibTeX key followed by the suffix ~.pdf~.  For example, if a BibTeX entry has the key ~Darwin1859~, bibtex-completion searches for ~Darwin1859.pdf~.
+Bibtex-completion assumes that the name of a PDF consists of the BibTeX key followed plus a user-defined suffix (~.pdf~ by default).  For example, if a BibTeX entry has the key ~Darwin1859~, bibtex-completion searches for ~Darwin1859.pdf~.
 
 If the BibTeX entries have a field that specifies the full path to the PDFs, that field can also be used.  For example, JabRef and Zotero store the location of PDFs in a field called ~File~:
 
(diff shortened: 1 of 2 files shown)


tmalsburg:
Are hyphens `-` allowed in Bibtex-keys?  I think yes, in which case there is room for ambiguities.  However, the chance of conflicts arising because of this is probably fairly low.  Other than that, this looks like a great extension.  Could you please add documentation in README?  Thank you!

ericdanan:
Yes hyphens are allowed, but I didn't understand what ambiguities could this induce?

Will update the readme.

tmalsburg:
Ambiguities: What if there is a bibtex key `Authors2016` and another bibtex key `Authors2016-supplement`? We would then list `Author2016-supplement.pdf` as a PDF for `Authors2016` even though it may not belong to this entry. As I said not very likely to happen but not impossible either. I'm probably in nitpicking-mode, so please feel free to ignore.

ericdanan:
OK, I didn't think about this. Personally I woudn't be too bothered by the case you describe, because `Authors2016-supplement` is presumably related to `Authors2016`. But for unrelated entries such as `Authors2016` and `Authors2016b`, I agree that would be annoying.

If we want to avoid this, then I think we need to check for each matching PDF filename of the form `Authors2016<suffix>.pdf` such that no "sub-suffix" of `<suffix>` matches, whether the bibliography has an entry with key `Authors2016<suffix>`. Now doing so by means of `bibtex-completion-get-entry` might be a bit slow (up to `0.25` seconds per check in my testing with `crypto.bib` which has ~40k entries). It seems to be much faster to call `cl-rassoc-if` on the candidates list (at most `0.04` seconds in my testing). But the problem is that since the candidates list is not stored anywhere, we first need to spend time rebuilding it form the cache (`0.45` seconds).

So it seems worthwile to introduce a variable `bibtex-completion-candidates` in which `helm-bibtex` or `ivy-bibtex` would store the candidates list so we wouldn't need to build this list a second time. (Actually I think this variable existed in the past and I wonder if it didn't disappear with a PR of mine...)

What do you think?

ericdanan:
Actually you can forget about the "sub-suffix" thing, in general we have to check all matching files except `Authors2016.pdf`. But if we can do this at `0.04` seconds per file it shouldn't be a problem.

tmalsburg:
Let's not worry about the ambiguity issue.  If it turns out to be a problem in practical cases, we can still add something.

If you add some text in the README file, I'll merge the PR.

ericdanan:
Done (I had to resolve a conflict in the readme.

I was not sure if it is worth keeping the sections on annotated pdfs and supplemental material, so I left them.

tmalsburg:
Thank you very much.  I think a lot of users will like these new features.

ericdanan:
You're welcome.

I noticed that you removed the readme section on annotated pdfs, but you kept the one on supplemental material that references it. I guess you want to remove it as well.

tmalsburg:
Ah thanks. Removed it. I think with the new workflow the additional actions wouldn't be needed anymore.

PR #220, opened 2017-12-14
fixes PDF loading issue issue where blank path could block later real path

ssfrr:
I'm using Zotero with the better-biblatex exporter, which apparently represents the PDF files as:

```
file = {/full/path/to/doc1.pdf;/full/path/to/doc2.pdf;/full/path/to/doc3.html}
```

When there's only one PDF this gets caught by [this line](https://github.com/tmalsburg/helm-bibtex/blob/84863a37695b786c6c6980a589f8ea282c385ab2/bibtex-completion.el#L638) and works fine. With multiple PDFs though, the path gets stored in `file-name` and `path` is `""`, and then the first element in `paths` is `""`, which exists so it gets assigned to `result`. Unfortunately it's a blank string so it gets killed. The problem is that there could be a valid path later on in the `paths` list. This PR tweaks the logic a little bit to get the first path that is both non-empty and exists.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -655,14 +655,19 @@ file is specified, or if the specified file does not exist, or if
                          (append
                           (list
                            path
+                           file-name
                            (f-join (f-root) path) ; Mendeley #105
                            (f-join (f-root) path file-name)) ; Mendeley #105
                           (--map (f-join it path)
                                  (-flatten bibtex-completion-library-path)) ; Jabref #100
                           (--map (f-join it path file-name)
                                  (-flatten bibtex-completion-library-path)))) ; Jabref #100
-           for result = (-first 'f-exists? paths)
-           if (not (s-blank-str? result)) collect result)))))))
+           for result = (-first (lambda (path)
+                                  (if (and (not (s-blank-str? path))
+                                           (f-exists? path))
+                                      path nil)) paths)
+           if result collect result)))))))
+
 
 (defun bibtex-completion-find-pdf-in-library (key-or-entry &optional find-additional)
   "Searches the directories in `bibtex-completion-library-path'


ssfrr:
Interestingly the original code also seems to work with windows paths (`C:\\Users\\...`) because it treats the `:` as a field separator, so `file-name` ends up as `C` and `path` is `\\Users\\...`. Then it gets `(f-root)` prepended, which is `c:/` on Windows.

tmalsburg:
Super sorry for the slow response.  The code for parsing the file field is fairly messy and brittle. Follows from the fact that there isn't really a standard format and every bibliography manager is doing something else.  Hugely annoying.

I could merge your PR but are we reasonably sure that this doesn't break other users' bibs?  Also, sidenote: separating PDFs with semicolon was not a great idea because semicolon is a perfectly valid character for file names.  Zotero probably escapes semicolons somehow, so your PR would have to deal with that special case as well.

ssfrr:
I agree regarding the annoyance of non-standard systems. I tried to make this PR as minimal as possible to minimize the risk of breaking something people were relying on. It makes 2 changes:

### Adds `file-name` to the list of candidate paths used to look for PDF files

`file-name` and `path` are the first and second colon-separated fields that some tools split their records into (I believe JabRef and Mendeley use something like `Description of the file:/full/path/to/file.pdf:PDF`). AFAICT Zotero (with the better-bibtex exporter) just puts the full path with no colon-separated fields. This means that `file-name` ends up being the path to the PDF and `path` ends up as `""`.

### Tweaks the candidate selection logic

The previous logic was:

1. collect a bunch of candidate paths
2. find the first candidate that satisfies `f-exists?`
3. use that result if it's not blank

the problem with that logic is that `""` satisfies `f-exists?` so it's selected in step 2, but then rejected in step 3. So no paths are found even if there are valid paths later in the candidate list. This PR changes the logic to:

1. collect a bunch of candidate path
2. use the first candidate that both satisfies `f-exists?` and is not blank

This should fix the zotero-better-bibtex case and fixes any case where a blank value shows up before the correct one.

The existing code already is splitting on semicolons [here](https://github.com/ssfrr/helm-bibtex/blob/04e3e3b67c4cbf93c9b11f68ade60420f4b48f58/bibtex-completion.el#L641), so I didn't touch that and if Zotero escapes them with something besides a backslash I think that would be a separate issue that would have already been broken.

I think this code should catch cases where the previous code failed, but I don't see that it could cause the code to fail to find files that it was finding before.

tmalsburg:
Okay, thanks for the explanation. All makes sense.  I'll merge and we'll see what happens.  Ultimately, I guess we need a plug-in system where people select from a number of styles and then the code just checks that style without trying all others.  Currently trying various different formats slows parsing down quite a bit.

tmalsburg:
p.s: Thanks for the PR!

PR #222, opened 2018-01-09
Fix byte compiler warnings

jabranham:
There is still one warning remaining after these fixes:

```
In toplevel form:
helm-bibtex.el:176:46:Warning: reference to free variable
    ‘helm-bibtex-local-bib’
```

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -39,6 +39,18 @@
 (require 'f)
 (require 'biblio)
 
+;; Silence byte-compiler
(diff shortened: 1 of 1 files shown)


tmalsburg:
Thank you! :-)

jabranham:
No problem. I can't see what `helm-bibtex-local-bib` is doing, and it doesn't seem to be defined on my Emacs. Can that part be removed?

tmalsburg:
Helm defines it on the fly.

PR #227, opened 2018-02-09
Fix byte compiler

jabranham:
- fixes byte-compile warning about an undefined variable
- simplifies a function call by removing if statement where the cond didn't matter
- documentation fixes, mostly for checkdoc

### DIFF
--- a/helm-bibtex.el
+++ b/helm-bibtex.el
@@ -131,13 +131,8 @@ When nil, the window will split below."
 ;; work-around.  See also this bug report:
 ;; http://debbugs.gnu.org/cgi/bugreport.cgi?bug=19395
 (defun helm-bibtex-window-width ()
-  (if (and (not (featurep 'xemacs))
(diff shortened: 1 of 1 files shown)


helm-bibtex.el
```diff
@@ -128,15 +131,11 @@ nil, the window will split below."
 ;; work-around.  See also this bug report:
 ;; http://debbugs.gnu.org/cgi/bugreport.cgi?bug=19395
 (defun helm-bibtex-window-width ()
-  (if (and (not (featurep 'xemacs))
```
tmalsburg: You are of course right that this doesn't do anything useful.  But I do have a faint memory that it used to do something. Hm ...

jabranham: Yeah, I figured that it used to do something. Hopefully it doesn't break anything. Easy to revert if people come complaining, though!

PR #231, opened 2018-03-14
enable use single notes.org in emacs 25.3.1

xingnix:
There is an error when using emacs 25.3.1 (other version not test). A single notes.org file works well after adding "(org-mode)"  before line 476.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -473,6 +473,7 @@ is the entry (only the fields listed above) as an alist."
     (when (and bibtex-completion-notes-path
                (f-file? bibtex-completion-notes-path))
       (with-temp-buffer
+	(org-mode)     ;;  need this to avoid error in emacs 25.3.1
         (insert-file-contents bibtex-completion-notes-path)
         (setq bibtex-completion-cached-notes-keys
               (let ((tree (org-element-parse-buffer 'headline)))


xingnix:
diff ~/Downloads/helm-bibtex-master/bibtex-completion.el  bibtex-completion.el
475a476
>         (org-mode)

tmalsburg:
What is the error that you're referring to?

xingnix:
When I use single note file and insert a note with heading line as:

```org
 *  Arya, M. C., & Semwal, A. (2017): Image classification based on image pixel value
  :PROPERTIES:
  :Custom_ID: 8276122
  :END:
  some notes here ...
```

then execute M-x helm-bibtex will produce a error message:

> wrong type argument: stringp, nill

and in buffer `*message*` there is 
> org-heading-components: Wrong type argument: stringp, nil

Tested in  Emacs 25.3.1 on  OpenSUSE Leap 42.3 / Tumbleweed

tmalsburg:
I'm asking because I'm also using a single notes file and don't get this error.  My suspicion is that there is something peculiar about your setup.  Before we make changes, we should figure out why this is only happening in your setup.  Have you tried to reproduce this issue with a minimal emacs setup (`emacs -Q` and so on)?

xingnix:
I tested with

> emacs -Q

and minimum pathes in the loadpath. 
The result is the same.

xingnix:
I installed helm-bibtex using elpa. I have considered the differences from the latest files in github (although there is very little difference) .

tmalsburg:
I suppose you mean MELPA?  To my knowledge the package is not available in ELPA.  One more question: Are you using org-ref?

xingnix:
Yes, it's melpa and I use org-ref.
At first I used org-ref and got the error message.  Then I found it load the function in helm-bibtex and the error message is same. I tried to solve the problem through ordinary customization (change value of symbols) and failed. At last I located the error function and got a solution by turning the tmp buffer  into org-mode.

jagrg:
I'm also seeing the error with emacs -Q. Which version of Org are you using? For some reason, in version 8 when we insert the file's content, org-heading-components is not aware of the major-mode. Maybe we could load Org as you suggest, but only for older versions.

```
(when (< (string-to-number (org-version)) 9)
  (org-mode))
```

tmalsburg:
I'm using version 9.1.6.

xingnix:
Mine is 8.2.10.

tmalsburg:
I guess it doesn't hurt loading org-mode unconditionally.

tmalsburg:
Thanks for the PR.

xingnix:
It's my honor to contribute.

PR #232, opened 2018-03-20
display ✎ for  entry key with dot ("." )

xingnix:
Correctly display pen symbol (✎) when there is dot ("." ) in a entry key ( like "123.456") .
It is not necessary to use function "regexp-quote"  because the function  "member" doesn't use regular expression.
With "regexp-quote" there is no (✎) because "." is  changed to "\\\\." but there is no such change in variable bibtex-completion-cached-notes-keys.


tmalsburg:
Makes total sense.  Thanks for the PR and sorry for the slow response.

xingnix:
I'm glad it helped.

PR #233, opened 2018-03-27
Fix trivial typos in README.

### LINKED ISSUE #219: readme file typos
(setq bibtex-completion-pdf-extenaion ".djvu")

If you store files in various formats, then you can specify a list instead of a single file type:

(setq bibtex-completion-pdf-extenaion '(".pdf" ".djvu"))

Shouldn't it be "extension" not extenaion?

tpapp:
Fixes #219.


tmalsburg:
Thank you!

PR #237, opened 2018-05-09
Add line to README about how to set up ivy

fbkarsdorp:
`bibtex-completion-candidates` creates search strings in which bibtex fields are ordered according to their order of appereance in a bibtex file. Unless one is very strict about ordering all entries in exactly the same way, this might lead to variable orderings. This is problematic in light of ivy's default regexp builder which expects all orderings to be the same. This problem can be resolved by using the (slightly slower) `ivy--regex-ignore-order` with ivy-bibtex.

Since ivy-bibtex is barely useable without this configuration, it's not really a matter of configuration but proper installation. As such, I added a line directly underneath `(autoload 'ivy-bibtex "ivy-bibtex" "" t)` in the Installation section.

### DIFF
--- a/README.org
+++ b/README.org
@@ -55,6 +55,12 @@ or
 
 #+BEGIN_SRC emacs-lisp
 (autoload 'ivy-bibtex "ivy-bibtex" "" t)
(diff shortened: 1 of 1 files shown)


tmalsburg:
Merged, thank you!!

PR #258, opened 2018-08-25
Postpone requiring org-element until needed

ericdanan:
This makes loading the bibiligraphy for the first time faster when org
is not loaded and notes are stored in multiple files.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -38,7 +38,6 @@
 (require 's)
 (require 'f)
 (require 'biblio)
-(require 'org-element)
 (require 'filenotify)
 (require 'subr-x)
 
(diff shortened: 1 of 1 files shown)


tmalsburg:
Out of curiosity: How much speedup do we get from this?  (I'm using a single notes file, so it doesn't affect me.)

tmalsburg:
I just tried `(require 'org-element)` with `emacs -Q` and it loaded instantly.  I wonder why it took so much longer in your experiment. Perhaps you're org sources aren't byte-compiled?

ericdanan:
My org sources are byte-compiled and it `(require 'org-element)` only take 0.4s with `emacs -Q` for me. The exra 2.2s come from various packages that I have automatically loaded after org-mode. So I guess the speedup will depend on the user's setup.

tmalsburg:
Okay, I merged the PR.  Thanks.

tmalsburg:
One more question: What is `(require org-element)` doing when it's already loaded? Are we now slowing down parsing for users of just one notes file?

ericdanan:
> What is (require org-element) doing when it's already loaded? 

Nothing (it does not reload the library).

> Are we now slowing down parsing for users of just one notes file (after the first time)?

Not noticeably, a second call to `(require 'org-element)` takes only 1.75e-05 second on my machine.

Issue #260, opened 2018-09-24
Use org-capture for bibtex-completion-notes?


leezu:
`org-capture` is widely used and "more powerful" than the current `bibtex-completion-edit-notes` setup. For example, the new notes can be organized in a datetree and the template code supports more things ([such as timestamps](https://github.com/tmalsburg/helm-bibtex/issues/151#issuecomment-256557236)).

I wrote below replacement for `bibtex-completion-edit-notes`. I don't suggest merging it just like this, but it may be a good starting point in case you are interested in migrating to some `org-capture` based setup. Compared to the original `bibtex-completion-edit-notes` it also uses indirect buffers instead of modifying the outline and narrow state of the original notes buffer. Only the single notes file case is supported. If a note already exist, the behavior is similar to the existing function (but using indirect buffer). For new notes, `org-capture-templates` is temporarily overwritten based on `bibtex-completion-notes-template-one-file` and then `org-capture` is invoked.

``` elisp
(defun bibtex-completion-edit-notes (keys)
       "Open the notes associated with the selected entry or create new via org-capture."
       (require 'org-capture)
       (dolist (key keys)
         (let* ((entry (bibtex-completion-get-entry key))
                (year (or (bibtex-completion-get-value "year" entry)
                          (car (split-string (bibtex-completion-get-value "date" entry "") "-"))))
                (entry (push (cons "year" year) entry))
                (buffer (generate-new-buffer-name "bibtex-notes")))
           (with-current-buffer (make-indirect-buffer  ;; TODO: buffer is never deleted
                                 (find-file-noselect bibtex-completion-notes-path)
                                 buffer t)  ;; clone t
             (widen)
             (outline-show-all)
             (goto-char (point-min))
             (if (re-search-forward (format bibtex-completion-notes-key-pattern (regexp-quote key)) nil t)
                                        ; Existing entry found:
                 (when (eq major-mode 'org-mode)
                   (org-narrow-to-subtree)
                   (goto-char (point-min))
                   (outline-show-all)
                   (switch-to-buffer-other-window buffer))
                                        ; Create a new entry:
               (let ((org-capture-templates
                      '(("bibtex" "helm-bibtex"
                         entry
                         (file+olp+datetree bibtex-completion-notes-path)
                         "%(s-format bibtex-completion-notes-template-one-file 'bibtex-completion-apa-get-value entry)"))))
                 (org-capture nil "bibtex")))))))
```

jagrg:
[...]

PR #263, opened 2018-10-17
 Use org-capture to create new notes

jagrg:
See https://github.com/tmalsburg/helm-bibtex/issues/260

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -574,7 +574,8 @@ is the entry (only the fields listed above) as an alist."
       (bibtex-completion-resolve-crossrefs files reparsed-files))
 
     ;; Finally return the list of candidates:
-    (message "Done (re)loading bibliography.")
+    (when reparsed-files
+      (message "Done reloading bibliography"))
     (nreverse
      (cl-loop
       for file in files
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -496,7 +496,8 @@ is the entry (only the fields listed above) as an alist."
     (when (and bibtex-completion-notes-path
                (f-file? bibtex-completion-notes-path))
       (with-temp-buffer
-	(org-mode)     ;;  need this to avoid error in emacs 25.3.1
+        (when (< (string-to-number (org-version)) 9)
+          (org-mode))     ; Avoid error in older versions of Org (see pull/231)
         (insert-file-contents bibtex-completion-notes-path)
         (setq bibtex-completion-cached-notes-keys
               (let ((tree (org-element-parse-buffer 'headline)))
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -50,6 +50,7 @@
 (declare-function org-find-property "org")
 (declare-function org-show-entry "org")
 (declare-function org-entry-get "org")
+(defvar org-capture-templates)
 
 (defgroup bibtex-completion nil
   "Helm plugin for searching entries in a BibTeX bibliography."
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -539,9 +539,10 @@ is the entry (only the fields listed above) as an alist."
     (when reparsed-files
 
       (with-temp-buffer
-        (insert-file-contents bibtex-completion-notes-path)
-        (let ((notes-hash (secure-hash 'sha256 (current-buffer))))
-          (setq bibtex-completion-notes-hash notes-hash)))
+        (when-let (notes-file bibtex-completion-notes-path)
+          (insert-file-contents notes-file)
+          (let ((notes-hash (secure-hash 'sha256 (current-buffer))))
+            (setq bibtex-completion-notes-hash notes-hash))))
 
       (cl-loop
        for file in files
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -1326,8 +1326,9 @@ If notes are not found, create one using org-capture (see
 	  ;; No notes found, so create one
           (progn
             (require 'org-capture)
-            (cl-pushnew bibtex-completion-capture-template
-                        org-capture-templates :test 'equal)
+            (unless (assoc-default "bibtex" org-capture-templates)
+              (cl-pushnew bibtex-completion-capture-template
+                          org-capture-templates :test 'equal))
 	    (org-capture nil "bibtex")
 	    ;; Move point past drawer and empty lines
 	    (org-end-of-meta-data t)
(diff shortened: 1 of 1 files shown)


tmalsburg:
Thank you for making this PR. A lot of things going on here.  I will need to make some time to study this.  Is this change going to introduce breaking chances for users of one or multiple notes files? What's the role of `bibtex-completion-notes-template-one-file` in this new implementation?  Is is redundant?  If I understand correctly it should serves the same purpose as the capture template no?  Sorry if I misunderstood something.

jagrg:
> Is this change going to introduce breaking chances for users of one or multiple notes files?

I don't think so. There should be no difference for multiple files users. For one file users, a few things changed. First, they should notice org-capture when creating new notes. Second, they should see the point after the properties drawer and not at the beginning of the heading. Third, they might have to change the bibtex-completion-capture-template variable if they want something different as far as org-capture goes.

I'm also hashing the notes file so that if the file hasn't changed since the last time it was parsed, bibtex-completion-candidates reuses the notes keys which should make helm-bibtex load faster.

> What's the role of bibtex-completion-notes-template-one-file in this new implementation? Is is redundant? If I understand correctly it should serves the same purpose as the capture template no? Sorry if I misunderstood something.

bibtex-completion-notes-template-one-file is what org-capture uses to create the notes template. The user could still do everything from the capture template but it would look something like this:

(setq bibtex-completion-capture-template
      '("bibtex" "helm-bibtex" entry (file+headline bibtex-completion-notes-path "New Notes")
	"%(bibtex-completion-clean-notes-string (s-format \"\n** ${author} (${year}) ${title}\n:PROPERTIES:\n:Custom_ID: ${=key=}\n:END:\n\n\"
	     'bibtex-completion-apa-get-value (bibtex-completion-get-entry entry)))"))

I think it may be easier for the user if we separate notes template from the capture template, WDYT? (cc @leezu)

bibtex-completion.el
```diff
@@ -1254,56 +1285,77 @@ line."
...
+If notes are not found, create one using org-capture (see
+`bibtex-completion-capture-template')."
+  (dolist (key keys)
+    (let ((entry (bibtex-completion-get-entry key)))
+      (if (not (assoc "=has-note=" entry))
+	  ;; No notes found, so create one
+          (progn
+            (require 'org-capture)
+            (unless (assoc-default "bibtex" org-capture-templates)
+              (cl-pushnew bibtex-completion-capture-template
```
leezu: Would it make sese to remove "bibtex" from `org-capture-templates` after completion of `bibtex-completion-edit-notes-one-file`? My concern is, that running `org-capture` separately from `bibtex-completion-edit-notes-one-file` should not show the "bibtex" action. (I haven't tested if this is the case with your code, but it seems to be the case?)

jagrg:
> Would it make sese to remove "bibtex" from `org-capture-templates` after completion of `bibtex-completion-edit-notes-one-file`? My concern is, that running `org-capture` separately from `bibtex-completion-edit-notes-one-file` should not show the "bibtex" action. (I haven't tested if this is the case with your code, but it seems to be the case?)

It's definitely not the case. At least I don't see the "bibtex" template show up, which I think is expected when we use a string and not a key as part of the template.

bibtex-completion.el
tmalsburg: Good point. I agree.

tmalsburg: Even better, could we just let-bind `org-capture-templates`?

tmalsburg:
@jarg, I haven't tested it but how is it possible that we permanently push a new template and then it doesn't show up?  I must be missing something.

jagrg:
On 17 Apr 2019, Titus von der Malsburg <notifications@github.com> wrote:

> @jarg, I haven't tested it but how is it possible that we permanently push a new template and then
> it doesn't show up? I must be missing something.

Try this:

(add-to-list 'org-capture-templates
	     `("bibtex" "Bibtex" entry (file ,(car org-agenda-files))
	       "* %?\n"))

Do you see anything different when you call org-capture interactively?

Now eval this:

(org-capture nil "bibtex")

tmalsburg:
Ah, I see what you man.  But I think "bibtex" as the key of the template is actually not allowed. See here: https://orgmode.org/manual/Template-elements.html#Template-elements

jagrg:
Are you sure? The docstring says:

ELisp programs can set KEYS to a string associated with a template
in `org-capture-templates'.  In this case, interactive selection
will be bypassed.

tmalsburg:
I think by "string" they mean the data type string:

> The keys that selects the template, as a string, characters only, for example ‘"a"’, 

Multiple characters are possible as well but there are constraints:

>  ‘"bt"’ for selection with two keys. When using several keys, keys using the same prefix key must be sequential in the list and preceded by a 2-element entry explaining the prefix key, for example:

tmalsburg:
Either way, I think we should perhaps not change the user's configuration in a way that may not be transparent to them.

bibtex-completion.el
leezu: I use let binding since opening https://github.com/tmalsburg/helm-bibtex/issues/260 and didn't observe any interference with "normal" org-capture use.

tmalsburg:
Sorry for letting this slide. Life is too busy and I don't find the time to study this PR more in depth.  Hopefully I will get to it soonish.  For now just one question: A solution relying on existing infrastructure (org-capture) would ideally involve fewer lines of code than something home-grown and would therefore be easier to maintain.  However, the present PR overall adds 52 lines.  Is there a way to make this feature more compact?  No need to write any code.  Just looking for ideas.  In general maintainability has become an issue with this software and it has prevented me from merging some other features as well.  If I had time, I'd heavily refactor the whole package.

jagrg:
I should've made smaller PRs. Sorry about that. Anyway, the idea is to use org-capture when creating new notes, so in theory we would need a new variable for defining the template, and a way to call org-capture programmatically when the Edit notes command is run. Also, I made the bibtex-completion-edit-notes smaller, in the hope of making it easier to maintain, by creating a separate function for the one file case and another one for the multiple files case. That's probably why the PR got so bloated. Anyway if you want it shouldn't be difficult to implement this feature from scratch.

tmalsburg:
Sorry for not getting back to you earlier. A solution using org-capture is the goal for the future.  Do you think #309 might be a good stepping stone?

jagrg:
Looks good AFAICT. It uses org-capture-fill-template to expand the
template which is nice. We should also consider replacing
bibtex-completion-notes-mode with org-capture-mode eventually.

tmalsburg:
Thanks @jagrg. I merged #309 with small modifications. There are still some good ideas in this PR that we don't have yet, but I will incorporate them individually (e.g. separate editing functions for one file vs. multiple files).

PR #267, opened 2018-10-25
Add `author-abbrev` format option

### LINKED ISSUE #266: Add abbreviated author list option for bibtex-completion-notes-template
For papers with 3 or more authors, it's common to abbreviate the author list to "First Author et al.". However, as far as I can tell, there is currently no way to do this in the `bibtex-completion-notes-template-*` variables.

I propose adding a `${author-abbrev}` (or similar) formatting option that behaves like the `%2a` format in `org-ref-reftex-format-citation`, such that one author is written as "Smith", two authors as "Smith and Jones", and three or more authors as "Smith et al.". Even better if the number of authors was customizable in the format (e.g. `${author-abbrev 3}` would list up to 3 authors in full, but 4 or more would be abbreviated to "et al.").

ashiklom:
For three or more authors, abbreviate to "First author et al."

This closes #266.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -137,12 +137,13 @@ suffix that is specified in `bibtex-completion-notes-extension'."
 
 (defcustom bibtex-completion-notes-template-multiple-files
   "#+TITLE: Notes on: ${author-or-editor} (${year}): ${title}\n\n"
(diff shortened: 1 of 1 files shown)


tmalsburg:
Looks good. Thank you!

PR #268, opened 2018-10-30
Do not require subr-x

jabranham:
We're not using any functions from subr-x

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -39,7 +39,6 @@
 (require 'f)
 (require 'biblio)
 (require 'filenotify)
-(require 'subr-x)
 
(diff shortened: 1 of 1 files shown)


tmalsburg:
This was a workaround for an issue with `if-let*` in Emacs 25 but it appears that `if-let*` isn't used anymore.  Thanks for catching this.

jabranham:
No problem.

If you do use if/when-let again, you can do `(eval-when-compile (require 'subr-x))` instead of requiring it at run time. That's the recommended way of using subr-x (documented in the commentary section of that file).

tmalsburg:
Ah, I see. Thanks for the info!

Issue #274, opened 2018-11-26
Open collection pdf containing article


mgttlinger:
If I have the pdf for a whole collection it would be nice to link all entries in that collection to open this pdf at the right point. I can set the pdf with the `file` field but then I have to manually search for the chapter/article. I can see that automatically linking this is probably out of scope given that the point in the pdf might not even be clear depending how the pages are numbered. Possibly different arguments for the viewer are required to open a pdf at a given page depending on the viewer used. However I think there is currently no way even to set this manually.

tmalsburg:
Good idea.  I think it will be difficult to add this as a standard feature because people use helm/ivy-bibtex with all kinds of PDF viewers some of which may allow this, others not.  However, I would be happy to add an example in the documentation showing how this could be implemented, e.g., with pdftools within Emacs.  Shouldn't be too difficult.

tmalsburg:
Oh, sorry, my message was ambiguous.  I meant I would be happy to add some code in the documentation if someone would provide it to me.  :)  I currently don't have the time to work on this.  Perhaps over the holidays.  Work is too hectic at this time.

mgttlinger:
Oh, OK yeah I misread that. I have a deadline on Monday. After that, I could look into that and see what I can come up with.

mgttlinger:
Ok, so how would I go about implementing this? 

My idea would be this:

1. Check in `bibtex-completion-open-pdf` if an additional key is in the bibtex entry specifying the starting page or a range of pages is specified in the entry.
2. Check after calling `bibtex-completion-open-pdf-function` if the buffer is in `pdf-view-mode`. (My understanding is that find file switches to that buffer allowing me to run commands in that buffer afterwards, right?)
3. Call `pdf-view-goto-label` with the starting page.

tmalsburg:
Sounds like a good plan but I think you have to actively switch to the PDF buffer, since the code for opening the PDF will probably be running in some helm (or ivy) buffer.

One issue: The pages of the PDF and the pages numbers often diverge since (for instance) title pages are counted in by the PDF viewer but not for the purposes of table of contents.  This means that `pdf-view-goto-label` might land us on a page that's earlier than the actual article or abstract.  Some PDF viewers are aware of the two separate ways to count pages but I'm not sure Emacs' pdf viewer is among them.

mgttlinger:
[...]

PR #276, opened 2018-12-18
Jumping to start page of article

mgttlinger:
As discussed in #274 pdfs can be opened at the "right" location for journal pdfs if pdf-tools is used.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -101,7 +101,7 @@ systems default viewer for PDFs is used."
   PDF if \"<key>.<extension\" exists."
   :group 'bibtex-completion
   :type 'boolean)
-  
+
 (defcustom bibtex-completion-pdf-symbol "⌘"
   "Symbol used to indicate that a PDF file is available for a
 publication.  This should be a single character."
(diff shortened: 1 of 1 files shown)


tmalsburg:
Thanks for the PR!  One question: Does this feature only work with pdftools or also with Emacs' native pdf-viewer?

tmalsburg:
Another question: If the entry is a journal article, the pages field may contain something like `146-189` but the PDF probably only contains a single article and the article will be on page one not 146.  Wouldn't the code in the PR try to jump to position 146?  Sorry if I misunderstood how it works.

mgttlinger:
> Does this feature only work with pdftools or also with Emacs' native pdf-viewer?

Currently only pdftools given that the functions I use are from pdftools. I could look into docview and see if there is something similar there.

> If the entry is a journal article, the pages field may contain something like 146-189 but the PDF probably only contains a single article and the article will be on page one not 146. Wouldn't the code in the PR try to jump to position 146?

That is what the function `bibtex-completion-pdf-find-offset` tries to handle. It checks how many pages the pdf has and if it has fewer pages than the start page of the article it jumps to page 1 instead. This is not the perfect or most advanced method to handle this but as an initial implementation, this seemed simple and working.

tmalsburg:
Sorry for the slow response.  I've looked at the PR again.  I suggest that we remove the code for the custom field startpage which probably no one else is using.  The two functions can then be combined into one.  Could you please also wrap long lines?  Thank you.

mgttlinger:
I'm not sure anymore if this really makes sense. I use this feature very rarely and given that it currently only works with pdf-tools its applicability is very limited.

tmalsburg:
Okay, in this case, I'll close this PR for now.  Thank you.

PR #277, opened 2019-01-04
Notes: narrow to the complete template instead of last org-heading

akoehn:
Before, a note template consisting of sub-headings would result in
only showing that sub-heading, e.g. using a template like this:

* Author (XXX): Title
** takeaways
** new questions
** additional literature

would only narrow to additional literature.  Now we narrow to the
complete template instead.

A side-effect is that for non-org templates the point is now at the
start of the template instead of the end.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -1329,13 +1329,14 @@ line."
             (goto-char (point-max))
             (save-excursion (insert (s-format bibtex-completion-notes-template-one-file
                                               'bibtex-completion-apa-get-value
-                                              entry))))
-          (when (eq major-mode 'org-mode)
-            (org-narrow-to-subtree)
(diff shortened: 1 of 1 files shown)


akoehn:
I thought about point positioning and would actually prefer if the point was always at the start but didn't want to also change that in this PR.

tmalsburg:
Thank you for this PR.  Makes perfect sense.  However, there is another PR (#263) in which we switch to org-capture as the underlying infrastructure.  I think this will also allow us to specify in the template where the cursor should be placed.  And correct narrowing should come for free if I'm not mistaken.

akoehn:
Yes, I saw that PR, but also that its last activity was in October.  This is just a minimal bugfix.

tmalsburg:
Merged, thank you!  (And neat idea to have predefined subsections.)

PR #281, opened 2019-02-14
Added IEEE searching to fallback options

mgttlinger:
biblio.el added IEEE searching capabilities recently. I added that to the fallback options.


tmalsburg:
Awesome. Thank you!

PR #288, opened 2019-04-25
Allow adding pdf from URL

ericdanan:


tmalsburg:
Looks super useful. Thank you!

PR #290, opened 2019-04-30
I add a function to show title in org-link-to-pdf scenario.

asvcc:
When you want to create a todo list about reading in org-mode, I think using PDF's title in the link is more intuitive.


tmalsburg:
This looks useful. Thank you.  Could you add some documentation in the README.org?  Otherwise it might be difficult for users to discover it.

tmalsburg:
Thank you.  Unfortunately, this got buried deep down in my to-do list. Sorry about that.

PR #291, opened 2019-05-09
allow underscore in citekeys

dlesl:
Hi! I don't know if I was doing something wrong or had something misconfigured, but I couldn't get `bibtex-completion-key-at-point` to recognise keys containing `_`. This fixes that, but there might be a better way? Maybe it would be OK to permanently modify the syntax table for `LaTeX-mode` instead.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -1454,7 +1454,11 @@ default to \"CUSTOM_ID\"). Otherwise, return nil."
              (let ((macro (reftex-what-macro 1)))
                (and (stringp (car macro))
                     (string-match "\\`\\\\cite\\|cite\\'" (car macro))
-                    (thing-at-point 'symbol)))))
+                    ;; allow '_' in citekeys
+                    (let ((temp-syn-table (make-syntax-table)))
+                      (modify-syntax-entry ?_ "_" temp-syn-table)
+                      (with-syntax-table temp-syn-table
+                        (thing-at-point 'symbol)))))))
       (and (eq major-mode 'org-mode)
            (let (key)
              (and (setq key (org-entry-get nil


tmalsburg:
Hey, sorry for the slow response and thanks for the PR. I was under the impression that underscores, due to their special status in LaTeX, were not allowed in BibTeX keys or at least strongly discouraged. For instance, when you print the BibTeX key in a document, this will produce an error because `_` is expected only in math mode.  Is there a specification that says underscores are okay in BibTeX keys?

dlesl:
Yes, after making the PR I started to notice that it's not so common to use them :). I started using them because Zotero's BibTeX exporter uses them by default, and I haven't had any problems getting LaTeX to accept them. With the reference styles common in my field the raw keys never appear in the final document, so I hadn't considered that! I also found [this stackexchange post](https://tex.stackexchange.com/questions/478896/zotero-and-bibtex-citation-keys-containing-underscore-are-unusable), there's a comment there which claims they work fine with both biblatex and bibtex.

tmalsburg:
Yes, I saw the Stackexchange post as well but also quite a few reports about problems with underscores in BibTeX keys. Perhaps it depends on what people are doing exactly in LaTeX.  If there is an official grammar that permits underscores in BibTeX keys or if it is sufficiently common to have underscores without major problems, we should support it somehow.

I'm not terribly familiar with syntax tables, but wouldn't we have to add the underscore to `_` (for symbol constituents according to the documentation of `modify-syntax-entry`) instead of `w`?

I would not permanently modify the syntax table because it's difficult to predict whether this breaks anything. (Actually, it likely breaks things related to `_` in math formulas where `x_i` would be treated as one constituent instead of multiple.

dlesl:
Hi, I think you're right that it should be in "_", I've changed this.

I really think this would be a good feature to have. I don't whether any official syntax exists, but I did find [this thread](https://emacs-orgmode.gnu.narkive.com/0YbW4krG/o-citation-syntax-underscore-must-be-allowed-in-cite-keys), which mentions that JabRef also supports underscores, and that Emacs's `bibtex.el` supports them explicitly (see `bibtex-reference-key`). I also recently finished writing a longish document using this format with biber/biblatex and encountered no issues. I noticed that during the first compilation pass (before running biber) the citation keys appear in the pdf, which suggests that biblatex escapes the underscore to make them printable.

tmalsburg:
Thanks for updating the PR.

PR #294, opened 2019-06-21
Change buffer selection method when exiting notes

JunnanZ:
Sometime I have multiple "notes.org" files open and the original method for selecting window will fail.

First time opening a pull request and pretty new to elisp. Hope this can help.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -1291,7 +1291,7 @@ line."
   (setq-local
    header-line-format nil)
   (save-buffer)
-  (let ((window (get-buffer-window (file-name-nondirectory bibtex-completion-notes-path))))
+  (let ((window (get-buffer-window (get-file-buffer bibtex-completion-notes-path))))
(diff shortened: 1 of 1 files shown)


tmalsburg:
Thank you for this PR.  Is the problem that this PR is solving described in an separate issue?  If not, and just to make sure that I understand what is going on, could you provide a minimal recipe that reproduces the problem?  Thank you!

JunnanZ:
For example, my default notes file is "notes.org" in some directory. Open that file and open another file with the same name in another directory. Then when you edit notes and call this function (C-c C-c), it will not delete the current window.

The reason is that `file-name-nondirectory` returns the file name "notes.org" but `get-buffer-window` can't get the correct buffer given this name. `get-file-buffer` is a more robust way to locate the buffer.

Hope it helps.

tmalsburg:
Ah, I see. Didn't think of the possibility that there are two buffers with files that have the same name.  Thanks for the fix!

Issue #295, opened 2019-06-27
handle LaTeX macros in bibtex-completion-insert-reference


plantarum:
Hi,

Many of my references contain italicized words (biological species names). eg.:

```
@article{durka_ea_2005,
  title        = {Molecular evidence for multiple introductions of garlic
                  mustard (\emph{Alliaria petiolata}, {Brassicaceae}) to {North
                  America}},
  ...
}
```

If I want to share the reference in an email, I insert it via ivy-bibtex. This removes all the braces, but leaves the macro, producing the following:

```
- Authors (2005). Molecular evidence
  for multiple introductions of garlic mustard (\emphAlliaria petiolata,
  Brassicaceae) to North America. 
  ...
```

This is a bit ugly. It would be better to leave the braces for the macro: `\emph{Alliaria petiolata}`, or to remove the macro as well as the braces `Alliaria petiolata`.

I think the code responsible is here:

https://github.com/tmalsburg/helm-bibtex/blob/e1391a4be629b506f06d678bcdfd17d290904482/bibtex-completion.el#L1120

I'm not sure if it would make sense to nest a `replace-regexp-in-string` form in there to remove any `\macro{` strings, or if there's another way to approach this?

Thanks for your time!

tmalsburg:
I'm not sure whether BibTeX entries should styling information.  The citation standards in my field are very strict and don't allow that.  But yes, I think it would probably be a good idea to remove LaTeX macros for the purpose of plain text references. If you'd like to prepare a PR, I'd be happy to give feedback.

tmalsburg:
I'm not sure whether BibTeX entries should styling information.  The citation standards in my field are very strict and don't allow that.  But yes, I think it would probably be a good idea to remove LaTeX macros for the purpose of plain text references. If you'd like to prepare a PR, I'd be happy to give feedback.

plantarum:
Ok, thanks. It looks like a simple fix, see what you think.

Best,

PR #296, opened 2019-07-03
remove formatting macros from plain text references

plantarum:
Addressing issue #295 

Added an additional replace-regexps-in-string clause to
bibtex-completion-apa-get-value: removes the pattern "\\MACRO{"" from
titles, where MACRO is arbitrary letters. This eliminates "\\emph{", and
any other macro forms that might be used to format text in a title.

Tested on examples from my .bib database - doesn't appear to break anything for me.


tmalsburg:
Great, thank you.

PR #300, opened 2019-08-07
Bugfix author-abbrev virtual field

ashiklom:
The original version was not treating "author-abbrev" as a virtual field, but rather as a literal bibtex field, so the "value" would never be retrieved from the bibtex entry. This fixes it so that "author-abbrev" is processed as a virtual field that asks for the author name.


tmalsburg:
Thanks for this fix.  (Wow, such a big diff for a small change.  A real downside of Lisp syntax.)

PR #301, opened 2019-08-08
Add a customisation for toggling of proactive reloading of bibliography.

mwlodarczak:


tmalsburg:
Thanks.  If someone loads helm-bibtex with the default setting for `bibtex-completion-watch-bibliography` and then decides that they don't want to proactively reload the bibliography, they have to restart Emacs with the new setting, right?

Issue #304, opened 2019-10-26
Better handling of pre/post-notes in bibtex-completion-format-citation-cite


gusbrs:
Currently, `bibtex-completion-format-citation-cite` handles prenotes and postnotes in citations by adding both arguments in case either of them is not empty, and none of them in case both are empty.

The citation syntax regarding pre- and postnotes is not quite universal, but a very general standard, shared by `biblatex` and `natbib` is that if there is only one optional argument, it is the postnote. If both optional arguments are supplied, the first is the prenote and the second is the postnote. So, one can specify `\cite[p.~45]{entrykey}` to get a postnote, but a lone prenote will require an empty second optional argument, as in `\cite[see][]{entrykey}`. This, no doubt, because the most common case is the use of the postnote.

So, currently `bibtex-completion-format-citation-cite` inserts both of them, even when there is only a postnote, thus including an unnecessary optional argument in the citation command, as in `\cite[][p.~45]{entrykey}`.

I kindly suggest the case of a postnote without a prenote be handled explicitly in `bibtex-completion-format-citation-cite` with the following change in the last part of the function:

```emacs-lisp
;; (if (and (string= "" prenote) (string= "" postnote))
;;     (format "\\%s{%s}" cite-command (s-join ", " keys))
;;   (format "\\%s[%s][%s]{%s}" cite-command prenote postnote (s-join ", " keys)))
(cond ((not (string= "" prenote))
       (format "\\%s[%s][%s]{%s}" cite-command prenote postnote (s-join ", " keys)))
      ((not (string= "" postnote))
       (format "\\%s[%s]{%s}" cite-command postnote (s-join ", " keys)))
      (t
       (format "\\%s{%s}" cite-command (s-join ", " keys))))
```
This will insert both optional arguments whenever a prenote is present, only one optional argument if a postnote is present without a prenote, and none of them if neither pre- nor postnote are supplied.

As noted above, the pre/postnote syntax is not quite universal (see https://tex.stackexchange.com/questions/471471/overwriting-harvard-style#comment1188095_471471). The standard LaTeX `\cite` command does not take a prenote. The `cite` package citation commands also not. `apacite` uses angled brackets for the prenote. And so on.

However, this suggestion gives a cleaner result for the cases where both pre- and postnote are used in the most common syntax (natbib and biblatex), and also handle the standard LaTeX syntax which does not take a prenote and would thus require manual editing in case a postnote was specified.

tmalsburg:
[...]

PR #305, opened 2019-10-28
Add optional arguments for LaTeX citation commands only where needed

gusbrs:
In attention to #304.

Inserts both optional arguments whenever a prenote is present, only one
optional argument if a postnote is present without a prenote, and none of them
if neither pre- nor postnote is supplied.


tmalsburg:
Merged.  Thank you very much!

gusbrs:
*I* thank *you*! Cheers!

PR #311, opened 2020-04-14
Isolate find-note-file lambda in prepare-entry

Zaeph:
Allows other packages to patch note-file-finding function without changing the
default behaviour of helm-bibtex.

This was prompted by #309.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -777,22 +777,23 @@ Returns nil if no PDF is found."
   (or (bibtex-completion-find-pdf-in-field key-or-entry)
       (bibtex-completion-find-pdf-in-library key-or-entry find-additional)))
 
-(defun bibtex-completion-find-note-file (entry-key)
-  "Find note file associated from BibTeX’s ENTRY-KEY.
-If the note file doesn’t exist, return nil."
+(defun bibtex-completion-find-note-file-default (entry-key)
+  "Find note file associated from BibTeX’s ENTRY-KEY in the default directory.
+The default directory is `bibtex-completion-notes-path'.  If the
+note file doesn’t exist, return nil."
   (and bibtex-completion-notes-path
        (f-directory? bibtex-completion-notes-path)
        (f-file? (f-join bibtex-completion-notes-path
                         (s-concat entry-key
                                   bibtex-completion-notes-extension)))))
 
-(defcustom bibtex-completion-find-note-file-fn
-  #'bibtex-completion-find-note-file
-  "Function to use to find note files.
-The function should accept one argument: the entry-key of the
+(defcustom bibtex-completion-find-note-file-fns
+  (list #'bibtex-completion-find-note-file-default)
+  "List of functions to use to find note files.
+The functions should accept one argument: the entry-key of the
 BibTeX entry."
   :group 'bibtex-completion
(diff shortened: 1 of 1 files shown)


tmalsburg:
Great, but I think we should allow `bibtex-completion-find-note-file-fn` to be a list of functions in case the user has notes in multiple places.  This approach would also mirror what I'm intending to implement for finding PDFs.  There, it will be useful as well to collect PDFs using multiple methods, e.g., by using the info from the entry's `file` field and by checking for CITKEY.pdf.

Zaeph:
>  But I think we should allow `bibtex-completion-find-note-file-fn` to be a list of functions in case the user has notes in multiple places.

I'm on it.

Zaeph:
> > But I think we should allow `bibtex-completion-find-note-file-fn` to be a list of functions in case the user has notes in multiple places.
> 
> I'm on it.

I've implemented it in 1083278.

Zaeph:
Note that we're only handling the *detection* of files here, not their *creation*.  To do that, the best way would be to implement fully-fledged `org-capture` templates, as discussed in #263.

Zaeph:
> Note that we're only handling the _detection_ of files here, not their _creation_. To do that, the best way would be to implement fully-fledged `org-capture` templates, as discussed in #263.

As it stands, the code could be modified [here](https://github.com/Zaeph/helm-bibtex/blob/10832788aebd7e8477a4ef0054cad4da7f05e50a/bibtex-completion.el#L1351-L1357), but I really don't think it'd be a good idea to play with it until we've refactored the templating with `org-capture`.

tmalsburg:
Merged into a feature branch because I'd like to handle one-file notes also via this infrastructure.

Zaeph:
> Merged into a feature branch because I'd like to handle one-file notes also via this infrastructure.

Sure, let me know if you need a hand!

tmalsburg:
Thanks, see here: 285f5189d36cccc0abafec9047fb203ee707c98c

I've made `bibtex-completion-find-note-functions` (formerly `bibtex-completion-find-note-file-fns`) a `defvar` instead of a `defcustom` because this is only the first step toward real modularity. There are multiple places that need to be changed before we can flexibly replace notes styles/plugins, first and foremost `bibtex-completion-edit-notes`.

Zaeph:
> Thanks, see here: [285f518](https://github.com/tmalsburg/helm-bibtex/commit/285f5189d36cccc0abafec9047fb203ee707c98c)
> 
> I've made `bibtex-completion-find-note-functions` (formerly `bibtex-completion-find-note-file-fns`) a `defvar` instead of a `defcustom` because this is only the first step toward real modularity.

That looks pretty good to me.  Sorry that I didn't modularise `bibtex-completion-find-note-one-file` whilst I was at it yesterday, it would have saved you some time.

> There are multiple places that need to be changed before we can flexibly replace notes styles/plugins, first and foremost `bibtex-completion-edit-notes`.

Yup, but we've made a good start here.  Good work! 🎉

tmalsburg:
>  Sorry that I didn't modularise bibtex-completion-find-note-one-file whilst I was at it yesterday, it would have saved you some time.

No worries. Your PR was a good start!

PR #318, opened 2020-04-25
Implement commands to only present candidates with notes

zaeph:
New commands:
- `helm-bibtex-with-notes` & `ivy-bibtex-with-notes` to only present candidates with notes.

I've modularised the code so that it can easily be expanded upon.  Rationale behind it was that I wanted a better way to access my Org-roam notes, and I figured this could be useful upstream.  Note that plays nicely with `bibtex-completion-find-note-functions`, so you could easily override its global value in a `let`-form to filter out items with a function.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -551,11 +551,10 @@ fields listed above) as an alist."
       for file in files
       append (cddr (assoc file bibtex-completion-cache))))))
 
-(defun bibtex-completion-candidates-with-notes ()
-  "Read the BibTeX files and return a list of candidates with notes.
+(defun bibtex-completion-candidates-with-notes (candidates)
+  "Return a filtered list of CANDIDATES without notes.
 See `bibtex-completion-candidates' for details."
-  (let ((candidates (bibtex-completion-candidates))
-        candidates-with-notes)
+  (let (candidates-with-notes)
     (dolist (candidate candidates candidates-with-notes)
       (when (assoc "=has-note=" candidate)
         (setq candidates-with-notes
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -562,25 +562,6 @@ See `bibtex-completion-candidates' for details."
               ;; nconc instead of push to avoid reversing
               (nconc candidates-with-notes (list candidate)))))))
 
-(defun bibtex-completion-get-candidates (&optional candidates)
-  "Retrieve the candidates to be used by the completion engine.
-
-By default, the candidates are retrieved with
-`bibtex-completion-candidates'.  If CANDIDATES is non-nil, it can
-either be a list of candidates as formatted by
-`bibtex-completion-candidates', or a function without any
-argument that returns such a list.
-
-See `bibtex-completion-candidates' for details."
-  (pcase candidates
-    ;; Default to `bibtex-completion-candidates'
-    ('() (bibtex-completion-candidates))
-    ((pred functionp) (funcall candidates))
-    ((pred listp) candidates)
-    (wrong-type (signal 'wrong-type-argument
-                        `((functionp listp)
-                          ,wrong-type)))))
-
 (defun bibtex-completion-resolve-crossrefs (files reparsed-files)
   "Expand all entries with fields from cross-referenced entries in FILES, assuming that only those files in REPARSED-FILES were reparsed whereas the other files in FILES were up-to-date."
   (cl-loop
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -551,16 +551,6 @@ fields listed above) as an alist."
       for file in files
       append (cddr (assoc file bibtex-completion-cache))))))
 
-(defun bibtex-completion-candidates-with-notes (candidates)
-  "Return a filtered list of CANDIDATES without notes.
-See `bibtex-completion-candidates' for details."
-  (let (candidates-with-notes)
-    (dolist (candidate candidates candidates-with-notes)
-      (when (assoc "=has-note=" candidate)
-        (setq candidates-with-notes
-              ;; nconc instead of push to avoid reversing
-              (nconc candidates-with-notes (list candidate)))))))
-
 (defun bibtex-completion-resolve-crossrefs (files reparsed-files)
   "Expand all entries with fields from cross-referenced entries in FILES, assuming that only those files in REPARSED-FILES were reparsed whereas the other files in FILES were up-to-date."
   (cl-loop
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -551,6 +551,17 @@ fields listed above) as an alist."
       for file in files
       append (cddr (assoc file bibtex-completion-cache))))))
 
+(defun bibtex-completion-candidates-with-notes ()
+  "Read the BibTeX files and return a list of candidates with notes.
+See `bibtex-completion-candidates' for details."
+  (let ((candidates (bibtex-completion-candidates))
+        candidates-with-notes)
+    (dolist (candidate candidates candidates-with-notes)
+      (when (assoc "=has-note=" candidate)
+        (setq candidates-with-notes
+              ;; nconc instead of push to avoid reversing
+              (nconc candidates-with-notes (list candidate)))))))
+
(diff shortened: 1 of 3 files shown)


tmalsburg:
If your symbol for notes (`bibtex-completion-notes-symbol`) is `⌘` for example, you can just search for that in helm or ivy to get entries with notes. If you want a separate helm command for serching just among those entries, you could let-bind the helm-bibtex default search string to be `bibtex-completion-notes-symbol`.  Would that work for your use case?  If yes, we could avoid duplicating search functionality in `bibtex-completion`.

zaeph:
> If your symbol for notes (`bibtex-completion-notes-symbol`) is `⌘` for example, you can just search for that in helm or ivy to get entries with notes. If you want a separate helm command for searching just among those entries, you could let-bind the helm-bibtex default search string to be `bibtex-completion-notes-symbol`. **Would that work for your use case?** If yes, we could avoid duplicating search functionality in `bibtex-completion` that is already present in helm/ivy. [emphasis added]

It could work, and I had thought of this at the start, but I find this solution a little fragile.  I've replaced the default `bibtex-completion-notes-symbol` with `N` early on because the symbol wasn't in my default font.

I thought this would be a nice opportunity to implement a function-filter for `bibtex-completion-candidates` in order to add flexibility.  I agree that the code isn't making a great demonstration out of it, but we could now have any function filter `bibtex-completion-candidates`, thereby giving us much finer control on the candidates we can display than what `ivy-` and `helm-bibtex` could achieve on their own.

(Although I'll readily admit that I might have had a revenge to take on that particular feature.  I'd tried to implement it at a much earlier point in my Elisp journey—2 years ago—and I'd been sorely defeated.  A *merge* might be due retribution for the trouble.)

tmalsburg:
Good point about note symbols like `N` not working well. Nonetheless, I think filtering is not in the scope of `bibtex-completion`. The purpose of bibtex-completion is merely to provide the candidates (and actions).  Filtering is a front-end matter.  How about having a command `helm-bibtex-with-notes` that wraps around `helm-bibtex` but which uses `cl-flet` to bind a filtered version of `bibtex-completion-candidates`?  This would mirror the approach used in `helm-bibtex-with-local-bibliography`.  Would that do the job?

zaeph:
> Good point about note symbols like `N` not working well. Nonetheless, I think filtering is not in the scope of `bibtex-completion`. The purpose of bibtex-completion is merely to provide the candidates (and actions). Filtering is a front-end matter. How about having a command `helm-bibtex-with-notes` that wraps around `helm-bibtex` but which uses `cl-flet` to bind a filtered version of `bibtex-completion-candidates`? This would mirror the approach used in `helm-bibtex-with-local-bibliography`. Would that do the job?

It could work, and thanks for making me discover `cl-flet`.  I'd have to consult my grimoires to know if this a good practice, and if there are caveats I should bear in mind.

I understand your concern about filtering being out of the scope `bibtex-completion`.  You mention `helm-bibtex-with-local-bibliography` as a command where I should get my inspiration, but the implementation for `helm-bibtex-with-notes` actually took most of its inspiration from it,  notably by adding an extra `CANDIDATES` argument to `helm-bibtex`.

Should I keep `bibtex-completion-candidates-with-notes` in `bibtex-completion.el`, then?  I know that you've said that `bibtex-completion` shouldn't handle filtering, but it's also the place where we keep most of the common actions between `helm-bibtex` and `ivy-bibtex`.

zaeph:
I'm not having much luck with preliminary testing:
```el
(defun helm-bibtex-with-notes (&optional arg)
  "Search BibTeX entries with notes.

With a prefix ARG the cache is invalidated and the bibliography
reread."
  (interactive "P")
  (cl-flet ((bibtex-completion-candidates () (bibtex-completion-candidates-with-notes)))
    (helm-bibtex arg nil nil)))
```

This does not shadow `bibtex-completion-candidates`.

zaeph:
This probably has something to do with lexical binding.  I'm investigating, but so far, not much success.  Investigating `cl-letf` now which does what is expected here:
```el
(defun zp/testing ()
  (message "Test!"))

(cl-letf (((symbol-function 'message) #'format))
  (zp/testing))
```

zaeph:
Ah, I know.  It's because `bibtex-completion-candidates-with-notes` makes use of `bibtex-completion-candidates` in its definition.  I should have listened more to the recursion warning!  It should be smooth sails from now.

zaeph:
1. 08942195125ac8c0625cce6b634132243d53217c modifies `bibtex-completion-candidates-with-notes` to *not* use `bibtex-completion-candidates` in its definition.  A new name might be warranted.
2. ad06e0899306f864e767a5f3ca5d0c2615915c58 reimplements the commands discussed above, but with `cl-letf`.  I'm a little worried about dynamically modifying a symbol in a `let`-form because an error in the execution would prevent the symbol from being reverted to its original assignment.  Not sure if there's a standard way to do this other than catching errors.

tmalsburg:
Just a quick note: I didn't forget this. Just super busy currently. Will get back to this soon, though.

zaeph:
> Just a quick note: I didn't forget this. Just super busy currently. Will get back to this soon, though.

Everything's fine, don't worry about it.  I'm running my branch and on my system, and it'll still be around when you get a little more time.  Good luck with everything!

tmalsburg:
What do you think about the following approach which doesn't require any changes in `bibtex-completion.el`:

``` elisp
(defun helm-bibtex-with-notes (&optional arg)
  "Search BibTeX entries with notes.

With a prefix ARG the cache is invalidated and the bibliography
reread."
  (interactive "P")
  (cl-letf* ((candidates (bibtex-completion-candidates))
             ((symbol-function 'bibtex-completion-candidates)
              (lambda ()
                (seq-filter
                 (lambda (entry) (assoc "=has-note=" entry))
                 candidates))))
    (helm-bibtex)))
```

>  I'm a little worried about dynamically modifying a symbol in a let-form because an error in the execution would prevent the symbol from being reverted to its original assignment.

Not sure I understand but it seems that `cl-letf` does revert values even after errors.  See this:

``` elisp
(setq test "123")
(cl-letf ((test "456"))
  (/ 1 0))
(message test)
```
This prints `"123"`.

zaeph:
> What do you think about the following approach which doesn't require any changes in `bibtex-completion.el`:

LGTM, and thanks for making me discover `seq-filter`.  Do you want me to commit the changes?

> Not sure I understand but it seems that `cl-letf` does revert values even after errors.

You're absolutely right.  I think the problem I had was due to an aborted `edebug` session at the end of which I surmised that it was `cl-letf`'s fault.

zaeph:
> Do you want me to commit the changes?

Did it in a13b7153df10ab3cb7fb6fd1da660f653b74f61d.

tmalsburg:
> Did it in a13b715.

Wow, that was quick.  Thank you!

> I surmised that it was cl-letf's fault.

Gotta say the descriptions of cl-letf that I saw were not clear in this regard and I was unsure as well.

Re seq.el, I think I should switch all list processing to seq.el one day to minimize dependencies of which we have probably too many.  But seq.el didn't exist yet when I started this project (I think).

tmalsburg:
Oops, there are two definitions of `helm-bibtex-with-notes` now.  Could you please remove the old one?  Thanks.

zaeph:
> Re seq.el, I think I should switch all list processing to seq.el one day to minimize dependencies of which we have probably too many. But seq.el didn't exist yet when I started this project (I think).

I saw that `seq-filter` was a generic function, and I automatically assumed that it must have quite a old function.

> Oops, there are two definitions of `helm-bibtex-with-notes` now. Could you please remove the old one? Thanks.

That's what I get for making speedy commits!  Fixed in 9a5d1f99a6c7bc5d6fece90a368b9440a59bd82e.

tmalsburg:
Merged, thanks again!

zaeph:
> Merged, thanks again!

Don't thank me, you did most of the work!

tmalsburg:
It's now possible to search for entries with notes or PDFs by entering `=has-note=` or `=has-pdf=`.  I wonder whether this renders the extra functions introduced by this PR obsolete. One alternative to removing them is to simplify them such that they use `=has-note=` as the default search string.  @zaeph, what do you think?

PR #320, opened 2020-05-01
Prevent error on nil with author-abbrev

zaeph:
Do not run `bibtex-completion-apa-format-authors-abbrev` is there is no author
on the entry.

-----

I've encountered the problem whilst testing zaeph/org-roam-bibtex/issues/20.

A better fix might be to implement an expansion for `author-or-editor-abbrev`.  I'm gonna try it now, and probably refactor the cond-check with `pcase`.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -1139,14 +1139,14 @@ The format depends on
   "Return FIELD or ENTRY formatted following the APA guidelines.
 Return DEFAULT if FIELD is not present in ENTRY."
   ;; Virtual fields:
-  (cond
-    ((string= field "author-or-editor")
+  (pcase field
+    ("author-or-editor"
      (let ((value (bibtex-completion-get-value "author" entry)))
        (if value
            (bibtex-completion-apa-format-authors value)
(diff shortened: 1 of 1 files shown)


tmalsburg:
Oh, thank you!

zaeph:
I've tried to implement a rudimentary `author-or-editor-abbrev`, hoping that I could just reuse `bibtex-completion-apa-format-authors-abbrev`, but it looks like you go `last, first` for all authors, and `first last` for all editors.

I think we'd want the two to be consistent for creating creating the headers of files.  So, have authors take precedence over editors when there are both, and format editors as authors (with `last, first` when there are only editors.  I'm pretty sure you're using this function for generating citations, though, so we'll have to keep this in mind.

zaeph:
Well, actually, we could just use `bibtex-completion-apa-format-authors-abbrev` and leave the rest intact.  Not sure why I havent thought of this before.  I'll make a new PR.

tmalsburg:
> but it looks like you go last, first for all authors, and first last for all editors.

This may be inconsistent but APA style mandates a different order for authors and editors.

PR #321, opened 2020-05-01
Implement author-or-editor-abbrev completion

zaeph:
Should solve the problems brought up in zaeph/org-roam-bibtex/issues/20 and #320 quite nicely.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -1192,119 +1192,81 @@ Return DEFAULT if FIELD is not present in ENTRY."
              (_ value))
          "")))))
 
-(defun bibtex-completion-apa-format-authors (value)
-  "Format author list in VALUE in APA style."
+(defun bibtex-completion-apa-format-authors (value &optional abbrev)
+  "Format author list in VALUE in APA style.
+When ABBREV is non-nil, format in abbreviated APA style instead."
   (cl-loop for a in (s-split " and " value t)
            if (s-index-of "{" a)
-             collect
-             (replace-regexp-in-string "[{}]" "" a)
-             into authors
+           collect
+           (replace-regexp-in-string "[{}]" "" a)
+           into authors
            else if (s-index-of "," a)
-             collect
-             (let ((p (s-split " *, *" a t)))
-               (concat
-                (car p) ", "
-                (s-join " " (-map (lambda (it) (concat (s-left 1 it) "."))
-                                  (s-split " " (cadr p))))))
-             into authors
+           collect
+           (let ((p (s-split " *, *" a t)))
+             (concat
+              (car p) ", "
+              (s-join " " (-map (lambda (it) (concat (s-left 1 it) "."))
+                                (s-split " " (cadr p))))))
+           into authors
            else
-             collect
-             (let ((p (s-split " " a t)))
-               (concat
-                (-last-item p) ", "
-                (s-join " " (-map (lambda (it) (concat (s-left 1 it) "."))
-                                  (-butlast p)))))
-             into authors
+           collect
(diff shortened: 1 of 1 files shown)


bibtex-completion.el
```diff
@@ -1140,11 +1140,14 @@ The format depends on
...
-       (bibtex-completion-apa-format-editors
-        (bibtex-completion-get-value "editor" entry))))
+    ((or "author-or-editor"
+         "author-or-editor-abbrev")
+     (let ((value (cl-some (lambda (field)
+                             (bibtex-completion-get-value field entry))
+                           (list "author" "editor"))))
+       (if (string-match-p "abbrev$" field)
+           (bibtex-completion-apa-format-authors value)
+         (bibtex-completion-apa-format-authors-abbrev value))))
```
zaeph, line 1150: Note that we're using the `format-authors` and `format-authors-abbrev`, since, if there aren't any author, we'd like to treat editors just as if they were themselves authors.

tmalsburg: This is neat but note that APA mandates that authors and editors are formatted differently (last first vs fist last). Perhaps the logic behind this is, that the order disambiguates authors from editors so that you can't get them confused.

zaeph: So, we could keep the general idea of having a new `author-or-editor-abbrev` expansion, but just adjust the order of names?  I could fix that.

tmalsburg: Yes. We could add virtual field "editor-abbrev" and then recurse when "author-or-editor-abbrev" is requested.

zaeph: Right, I'll need to modify `bibtex-completion-apa-format-authors-abbrev` to do our bidding, then.  I'm a bit short on time right now, but I'll get to it within the fortnight.

tmalsburg: I'd create a separate function `bibtex-completion-apa-firmat-editors-abbrev` for that purpose.  Otherwise the naming becomes misleading. But these two functions can of course share common code.

zaeph: Oh yeah, sorry, I meant to say 'modify _and_ create a new function', but we're on the same page.

zaeph: Prototyped in fb5b30195aa6540b1ac78cd4922fc1e165c6e375.

zaeph:
fb5b30195aa6540b1ac78cd4922fc1e165c6e375 has a working prototype.  I'm going to modularise the code so that we don't have to repeat huge swaths of code, now.

zaeph:
>  I'm going to modularise the code so that we don't have to repeat huge swaths of code, now.

Done in 0339f478d20cba708b43b803fb6fc1d104271f7f.

bibtex-completion.el
```diff
@@ -1184,93 +1193,81 @@ Return DEFAULT if FIELD is not present in ENTRY."
              (_ value))
          "")))))
 
-(defun bibtex-completion-apa-format-authors (value)
-  "Format author list in VALUE in APA style."
+(defun bibtex-completion-apa-format-authors (value &optional abbrev)
+  "Format author list in VALUE in APA style.
+When ABBREV is non-nil, format in abbreviated APA style instead."
   (cl-loop for a in (s-split " and " value t)
```
zaeph, line 1199: Uhm, my `lisp-indent-function` seems to have borked your indent.  I don't know much CL, so I'm not sure which one is standard, but I assume yours was.

tmalsburg: I think mine was non-standard but I used it nonetheless because I think it's more readable. But no big deal.

tmalsburg:
Merge.  Thank you.

PR #323, opened 2020-05-02
Use _ instead of t with s-join

zaeph:
Got a warning that it was deprecated when loading bibtex-completion.  I don’t
use s-join too much, so you might want to verify that everything works as
expected.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -1177,7 +1177,7 @@ Return DEFAULT if FIELD is not present in ENTRY."
              ("booktitle" value)
              ;; Maintain the punctuation and capitalization that is used by
              ;; the journal in its title.
-             ("pages" (s-join "–" (s-split "[^0-9]+" value t)))
+             ("pages" (s-join "–" (s-split "[^0-9]+" value _)))
              ("doi" (s-concat " http://dx.doi.org/" value))
              ("year" (or value
                          (car (split-string (bibtex-completion-get-value "date" entry "") "-"))))


zaeph:
Well, actually, it doesn't work.  I'll investigate.

zaeph:
So this seems to be due to the `pcase` forms that wraps around this line:
https://github.com/tmalsburg/helm-bibtex/blob/16b113692966da9bbe1f9f0ed8dcbfd2db61a7ba/bibtex-completion.el#L1180

I'm not sure why we get a warning for this because the `t` is not in the condition-part of the `pcase`, so that's a bit strange.

tmalsburg:
I don't seem to be getting this warning.  What do I have to do to get it?

zaeph:
> I don't seem to be getting this warning. What do I have to do to get it?

Try running `flycheck` in `bibtex-completion.el`, and check this line:
https://github.com/tmalsburg/helm-bibtex/blob/16b113692966da9bbe1f9f0ed8dcbfd2db61a7ba/bibtex-completion.el#L1180

tmalsburg:
Just ran flycheck, no warning on that line.  Hm.

zaeph:
> Just ran flycheck, no warning on that line. Hm.

'Hm' indeed.

screenshot-02f1Ij9GUu

tmalsburg:
Wierd. But I checked, we can't just supply `_` to s-split because that isn't non-nil, it's void.  I think this is a flycheck bug and needs to be fixed there.

zaeph:
> Wierd. But I checked, we can't just supply `_` to s-split because that isn't non-nil, it's void. I think this is a flycheck bug and needs to be fixed there.

Agreed.  Sorry for the scare!

PR #336, opened 2020-08-29
Add a user option to control whether to shorten authors or not

kisaragi-hiu:
###### Rationale
I personally prefer to not shorten the author name in my bibtex entries, because not all names end with a family name. Additionally, sometimes it makes the most sense to attribute a work to an entire group (`author = {Valve Corporation}`, for instance), and shortening these names can remove too much information.

While this can be worked around with a

```elisp
(defalias 'bibtex-completion-shorten-authors #'identity)
```

, it is more discoverable for others who prefer to not shorten names if it's actually available as a user option.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -339,6 +339,11 @@ editor names."
   :group 'bibtex-completion
   :type '(alist :key-type symbol :value-type string))
 
+(defcustom bibtex-completion-shorten-authors t
+  "If t, bibtex-completion shortens author names."
+  :group 'bibtex-completion
+  :type 'boolean)
+
 (defvar bibtex-completion-cross-referenced-entry-types
   '("proceedings" "mvproceedings" "book" "mvbook" "collection" "mvcollection")
   "The list of potentially cross-referenced entry types (in lowercase).
@@ -891,7 +896,7 @@ governed by the variable `bibtex-completion-display-formats'."
 
 (defun bibtex-completion-shorten-authors (authors)
   "Return a comma-separated list of the surnames in AUTHORS."
-  (if authors
+  (if (and authors bibtex-completion-shorten-authors)
       (cl-loop for a in (s-split " and " authors)
                for p = (s-split "," a t)
                for sep = "" then ", "


tmalsburg:
If you have a name that you don't want to have shortened, I think you can simply enclose it in curly braces (`author = {{Valve Cooperation}}`).  Could you please check? My memory is faint.

In general, I think this should not be a user option because I'd consider a shortened company name a bug.

kisaragi-hiu:
The name is still shortened with an extra level of curly braces.

This is because:

- While `bibtex-completion-get-value` does preserve the second level of braces,
- `bibtex-completion-format-entry` always passes the value of `authors` to `bibtex-completion-shorten-authors`, and
- `bibtex-completion-shorten-authors` does not take the braces into account.

tmalsburg:
Thank you.  In the case of company names my impression is, that we might need to change `bibtex-completion-shorten-authors' such that it retrieves the raw value of the author field and interprets it in a way that respects the braces.

Could you please give me more details about the case where the family name is not the last token?  If the last token is not the family name, how does LaTeX recognize this?

kisaragi-hiu:
I'm not sure what "the last token" means; assuming that you meant when the last word isn't the family name: [name order is dependent on culture](https://en.wikipedia.org/wiki/Given_name#Name_order) and is a mess full of special cases. East Asian countries put family names in front, for instance.

As far as I know, people just use their family names as the last name regardless of their native or preferred name order, to allow LaTeX styles to recognize it. But IMO as this is only a display feature in bibtex-completion, we don't have to force this convention.

That said, I think this PR isn't the best implementation for this. It's probably better to allow the user to choose a formatter for names or define their own. For instance, this is my current config:

```elisp
(defun kisaragi/bibtex-shorten-names (names)
  "Return a comma-separated list of names in NAMES (a string)."
  (->> names
       (s-replace " and others" " et al.")
       (s-replace " and " ", ")))
(defalias 'bibtex-completion-shorten-authors
  #'kisaragi/bibtex-shorten-names)
```

tmalsburg:
I've created a new issue for this #345.

Regarding names with a different order of given and family name, I have to confess that I don't understand the problem.  Bibtex / LaTeX expect names in the order `first last` or `last, first` to be able to typeset references correctly. None of these order is natural for names from cultures with reverse order but if the names are given in the order expected by BibTeX / LaTeX, helm-bibtex will display them correctly I think.

PR #338, opened 2020-09-01
Fix entries overflowing if they include wide characters

kisaragi-hiu:
Using eg. "${title:*}" in `bibtex-completion-display-formats` should truncate that field to the remaining window width, but previously that calculation fails to account for characters wider than one column, like CJK characters or tabstop.

This PR fixes that by using `string-width` (added in Emacs 20) instead of `length` to determine the displayed width of the string.

Demonstration:

```elisp
  '((t . "${year:4}\t${=has-pdf=:1}${=has-note=:1} ${=type=:10}\t${author:20}\t${title:*}\t${keywords:20}"))
```

Before (note how entries are wrapping because the remaining window width has been overestimated):

before

Fixed:

fixed

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -417,7 +417,7 @@ Also sets `bibtex-completion-display-formats-internal'."
                   (let* ((format-string (cdr format))
                          (fields-width 0)
                          (string-width
-                          (length
+                          (string-width
                            (s-format format-string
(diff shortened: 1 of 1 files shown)


tmalsburg:
Didn't know about `string-width` (despite the fact that I used the variable name `string-width`, lol). Thank you very much!!

PR #349, opened 2021-01-08
parcify the warning of non-used argument `default'

ShuguangSun:
The argument `default` is not used in `bibtex-completion-apa-get-value`.


tmalsburg:
Thanks.  Not sure whether default values are ever used tbh.

Issue #353, opened 2021-01-22
[Feature/WIP] Add selectrum et al support


mtreca:
First of all thanks for this great package.

I recently moved to the [selectrum](https://github.com/raxod502/selectrum) completing framework, which plays very nicely with a certain number of other utilities such as [marginalia](https://github.com/minad/marginalia) for minibuffer annotations and [embark](https://github.com/oantolin/embark) for minibuffer actions.

Having a `selectrum-bibtex` frontend leveraging these tools would be great. I started working on a POC, I will open a PR when something usable comes out of it.

bdarcus:
Yes please!

mtreca:
Awesome. I managed to port most of the features from ivy-bibtex quite easily. Working on marginalia support now!

minad:
@mtreca I am not sure if you are using my [Consult](https://github.com/minad/consult/) package, but I wonder if you could do the selectrum-bibtex port in the spirit of Consult by using `completing-read` instead of Selectrum-specific APIs? This way one would get compatibility with other completion systems like default completion, Icomplete-vertical and Selectrum.

mtreca:
I am actually using your `consult--read` helper function in my script, which I guess is `completing-read` friendly. I will look into it!

minad:
@mtreca Yes, `consult--read` is using `completing-read`. It is okay to use `consult--read` if

* You need the special functionality provided by `consult--read`.
* You are fine with making your package depend on Consult. 
* You are fine with possibly being affected by breakage due to changes of `consult--read` (Hopefully such changes will be rare)

mtreca:
While I think that most selectrum users will have consult,  marginalia et al installed it might be better to use `completing-read` to limit dependencies. Embark will be necessary however since selectrum-bibtex will have to provide multiple actions during completion.

tmalsburg:
Hi all, I'm not familiar with selectrum but in principle I'm happy to include support for frameworks other than helm and ivy. Note however that bibtex-completion (the backend) is available as a separate MELPA package, so you could host selectrum support also in a separate repository.  I don't know enough about what you're planning to be able to say whether that makes more sense than including selectrum support here.  The ivy and helm front-ends are sufficiently similar to benefit from being maintained by the same person in the same repo but the selectrum front-end may be different.

minad:
[...]

PR #355, opened 2021-02-01
[WIP] Add embark-bibtex frontend.

mtreca:
Following issue #353, I worked on a prototype frontend for bibtex-completion leveraging the `embark` and `marginalia` packages.

What changed since last iteration:

- Remove `consult` dependency.
- Add rudimentary marginalia annotations.

What it looks like:

2021-02-01-211516_1920x1080_scrot

What is missing (a lot):

- The front-facing functions do not support the optional args from `ivy-bibtex` (will add later once the prototype becomes more usable).
- No simple mechanism to display the notes and pdf icons so far (I did not look very hard at this issue yet, though).
- The way marginalia annotations are built is very hacky so far. It does not rely on any of the tools or variables provided by bibtex-completion (such as `bibtex-completion-format-entry`). From what I gather, this is because the default completion mechanism cannot rely on fancier display functions like `ivy-set-display-transformer`. This means that I have to parse the completion options to only fetch the entry keys, pass them to `completing-read`, and re-create annotations on top of that. It would be much nicer to reuse the bibtex-completion built-in tools.
- Linked to the point above, annotation alignment is kind of tricky since I cannot rely on `bibtex-completion-format-entry`.

This is of course far from being PR-ready, but I would like to make this front-end a more collaborative effort, especially from people knowing both marginalia and bibtex-completion more than me.

### DIFF
--- a/embark-bibtex.el
+++ b/embark-bibtex.el
@@ -1,0 +1,150 @@
+;;; embark-bibtex.el --- A bibliography manager based on Embark
+
+;; Author: Maxime Tréca <maxime.treca@gmail.com>
+;; Maintainer: Titus von der Malsburg <malsburg@posteo.de>
+;; URL: https://github.com/tmalsburg/embark-bibtex
+;; Version: 1.0.0
+;; Package-Requires: ((bibtex-completion "1.0.0") (embark "0.10")
+;; (cl-lib "0.5") (marginalia "0.2"))
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; A BibTeX bibliography manager based on built-in completion, embark
+;; completion actions and the bibtex-completion backend.
+;;
+;; Key features:
+;; - Quick access to your bibliography from within Emacs
+;; - Tightly integrated workflows
+;; - Provides instant search results as you type
+;; - Powerful search expressions
+;; - Open the PDFs, URLs, or DOIs associated with an entry
+;; - Insert LaTeX cite commands, Ebib links, or Pandoc citations,
+;;   BibTeX entries, or plain text references at point, attach PDFs to
+;;   emails
+;; - Attach notes to publications
+;;
+;; Install:
+;;
+;;   Put this file in a directory included in your load path or
+;;   install embark-bibtex from MELPA (preferred).  Then add the
+;;   following in your Emacs startup file:
+;;
+;;     (require 'embark-bibtex)
+;;
+;;   Alternatively, you can use autoload:
+;;
+;;     (autoload 'embark-bibtex "embark-bibtex" "" t)
+;;
+;;   Requirements are parsebib, embark, s, dash, and f.  The easiest way
+;;   to install these packages is through MELPA.
+;;
+;;   Let embark-bibtex know where it can find your bibliography by
+;;   setting the variable `bibtex-completion-bibliography'.  See the
+;;   manual for more details:
+;;
+;;     https://github.com/tmalsburg/helm-bibtex/blob/master/README.embark-bibtex.org
+;;
+;; Usage:
+;;
+;;    Do M-x embark-bibtex and start typing a search query when prompted.
+
+;;; Code:
+
+(require 'embark)
+(require 'marginalia)
+(require 'bibtex-completion)
+
+(defcustom embark-bibtex-default-action 'embark-bibtex-open-any
+  "The default action for the `embark-bibtex` command."
+  :group 'bibtex-completion
+  :type 'function)
+
+(defmacro embark-bibtex-embarkify-action (action name)
+  "Wraps the function ACTION in another function named NAME which
+extracts the key from the candidate selected in embark and
+passes it to ACTION."
+  `(defun ,name (cand)
+     (interactive "s ")
+     (,action (list cand))))
+
+(embark-bibtex-embarkify-action bibtex-completion-open-any embark-bibtex-open-any)
+(embark-bibtex-embarkify-action bibtex-completion-open-pdf embark-bibtex-open-pdf)
+(embark-bibtex-embarkify-action bibtex-completion-open-url-or-doi embark-bibtex-open-url-or-doi)
+(embark-bibtex-embarkify-action bibtex-completion-insert-citation embark-bibtex-insert-citation)
+(embark-bibtex-embarkify-action bibtex-completion-insert-reference embark-bibtex-insert-reference)
+(embark-bibtex-embarkify-action bibtex-completion-insert-key embark-bibtex-insert-key)
+(embark-bibtex-embarkify-action bibtex-completion-insert-bibtex embark-bibtex-insert-bibtex)
+(embark-bibtex-embarkify-action bibtex-completion-add-PDF-attachment embark-bibtex-add-PDF-attachment)
+(embark-bibtex-embarkify-action bibtex-completion-edit-notes embark-bibtex-edit-notes)
+(embark-bibtex-embarkify-action bibtex-completion-show-entry embark-bibtex-show-entry)
+(embark-bibtex-embarkify-action bibtex-completion-add-pdf-to-library embark-bibtex-add-pdf-to-library)
+
+(defun embark-bibtex--get-candidates ()
+  "Return all keys from bibtex-completion-candidates."
+  (mapcar (lambda (cand)
+            (cdr (assoc "=key=" cand)))
+          (bibtex-completion-candidates)))
+
+;;;###autoload
+(defun embark-bibtex (bib-entry)
+  "Search BibTeX entries using `completing-read' and embark
+actions."
+  (interactive
+   (list
+    (completing-read
+     "BibTeX entries: "
+     (lambda (string predicate action)
+       (if (eq action 'metadata)
+           '(metadata (category . bibtex))
+         (complete-with-action action (embark-bibtex--get-candidates) string predicate))))))
+  (apply embark-bibtex-default-action (list bib-entry)))
+
+(defun marginalia-annotate-bibtex (bib-key)
+  "Return completion annotations for a bibtex key."
+  (let ((bib-entry (bibtex-completion-get-entry bib-key)))
+    (concat
+     (marginalia--fields
+      ((cdr (assoc "title" bib-entry)) :width 70 :truncate 70)
+      ((cdr (assoc "author" bib-entry)) :width 30 :truncate 30)
+      ((cdr (assoc "year" bib-entry))  :width 5 :truncate 5)))))
+
+(add-to-list 'marginalia-annotators-light '(bibtex . marginalia-annotate-bibtex))
+(add-to-list 'marginalia-annotators-heavy '(bibtex . marginalia-annotate-bibtex))
+
+(embark-define-keymap embark-bibtex-map
+  "Keymap for actions for bibtex."
+  ("p" embark-bibtex-open-pdf)
+  ("u" embark-bibtex-open-url-or-doi)
+  ("c" embark-bibtex-insert-citation)
+  ("r" embark-bibtex-insert-reference)
+  ("k" embark-bibtex-insert-key)
+  ("b" embark-bibtex-insert-bibtex)
+  ("a" embark-bibtex-add-PDF-attachment)
+  ("e" embark-bibtex-edit-notes)
+  ("s" embark-bibtex-show-entry)
+  ("l" embark-bibtex-add-pdf-to-library))
+
+(add-to-list 'embark-keymap-alist '(bib . embark-bibtex-map))
+
+(provide 'embark-bibtex)
+
+;; Local Variables:
+;; byte-compile-warnings: (not cl-functions obsolete)
+;; coding: utf-8
+;; indent-tabs-mode: nil
+;; End:
+
+;;; embark-bibtex.el ends here


bdarcus:
Seems this will be a great test for embark and marginalia!

FWIW, my biggest needs are elegant support for citation key insertion in org and markdown for document authoring, and ability to integrate with org-roam-bibtex (which takes over note-taking for bib notes for use with org-roam).

minad:
Note that there is also no reason to depend on Marginalia if you provide your own annotation function. You can provide your own annotation-function in completing-read as part of the metadata. Marginalia is really meant to enhance existing commands (switch-to-buffer, find-file, etc.), not to be used in scenarios like here. I think I've mentioned that before. The only dependency you really need is Embark for action support.

minad:
If you want to see prefix annotations (icons for example), you should use affixation-function instead of annotation-function. But this is only supported by Selectrum and Emacs 28 default completion.

bdarcus:
Out of my depth, but wouldn't `bibtex-completion-display-formats` cover this?

minad:
@bdarcus Yes, by providing your own annotation function you can do your own formatting in the way you want it. This is better than using the marginalia specific internals.

minad:
Maybe one more suggestion - you can also remove the Embark dependency and just provide the keymap and the actions as command. Then this keymap should be bound via Embark to the completion category in the init.el use-package definition. But this way you already get commands out of all the actions if you only have bibtex-completion+this-new-package. And then my point was that the stuff should just be moved into the bibtex-completion package altogether :)

bdarcus:
> Yes, by providing your own annotation function you can do your own formatting in the way you want it.

Well, and it provides default formatting; right?

tmalsburg:
Looks promising!

> No simple mechanism to display the notes and pdf icons so far (I did not look very hard at this issue yet, though).

Have a look at `bibtex-completion-format-entry`.

embark-bibtex.el
```diff
@@ -0,0 +1,150 @@
...
+(embark-bibtex-embarkify-action bibtex-completion-add-pdf-to-library embark-bibtex-add-pdf-to-library)
+
+(defun embark-bibtex--get-candidates ()
+  "Return all keys from bibtex-completion-candidates."
+  (mapcar (lambda (cand)
+            (cdr (assoc "=key=" cand)))
+          (bibtex-completion-candidates)))
+
+;;;###autoload
+(defun embark-bibtex (bib-entry)
```
bdarcus, line 102:
```suggestion
(defun embark-bibtex (local-bib)
```
Minor thing here, but `bib-entry` suggests a single item, rather than the full collection. This suggests uses the same name as `ivy-bibtex` and `helm-bibtex`.

mtreca:
Thanks a lot for the feedback everyone!

@minad I did not know you could provide your own annotation function in `completing-read`. That is great news as I can indeed remove the `marginalia` dependency for now. As for removing the embark dependency, I am not sure I understand what you mean yet, but thanks for the advice. I will likely ping you later when I will tackle this, if you don't mind.

@bdarcus @tmalsburg Thanks for the remarks.

I will take all of your feedback into account for the next iteration.

mtreca:
One limitation of marginalia and providing completion annotations is that the annotations are not matchable when selecting a candidate. For instance, if one of the entries is displayed as

```
⌘ 2006 Yu, X-H and Recker, Wilfred W Stochastic adaptive control model for traffic signal systems article yu2006stochastic
```

`ivy-bibtex` (and I guess `helm-bibtex`) allow for matching parts of the title of authors. Providing annotations for standard `completing-read` (through annotation functions or `marginalia`) only allow to match the article key (in this case `yu2006stochastic`).

I see this as a pretty big limitation of the `embark-bibtex` frontend since fuzzy searching article titles or authors is extremely helpful IMO.

An option would be to wrap completion candidates using cons cells, as described by John Kitchin in [this answer](https://emacs.stackexchange.com/questions/42099/how-can-i-ask-a-user-to-choose-from-a-list-of-options-while-showing-her-a-long).

bdarcus:
> Providing annotations for standard `completing-read` (through annotation functions or `marginalia`) only allow to match the article key (in this case `yu2006stochastic`).
> 
> I see this as a pretty big limitation of the `embark-bibtex` frontend since fuzzy searching article titles or authors is extremely helpful IMO.

Indeed; if this is the case, I would find it unusable. The whole point is, if I have 1000 items in my db, I don't know the key. So I type in parts of authors and titles to narrow it.

But I thought bibtex-complete presented an alist?

minad:
> I see this as a pretty big limitation of the embark-bibtex frontend since fuzzy searching article titles or authors is extremely helpful IMO.

There is no problem actually. Just make everything you want to match on part of the candidate string. Annotations are only auxiliary information which is per definition by Emacs not supposed to be matched on. Maybe annotations are not even necessary since you want to be able to match on everything?

bdarcus:
> Just make everything you want to match on part of the candidate string.

So @mtreca - can you not then just use `bibtex-completion-format-entry` to match on, which then allows the user also to configure the details of that via `bibtex-completion-display-formats`?

mtreca:
@minad Yes this is what I am going for. it is actually easier than what I thought it was going to be. This snippet does proper completion matching, leverages built-in formatting functions from `bibtex-completion`, and does not use `marginalia`!

```elisp
(let ((candidates
       (mapcar
        (lambda (cand)
          (cons (bibtex-completion-format-entry cand 188)
                (cdr (assoc "=key=" cand))))
        (bibtex-completion-candidates))))
  (cdr (assoc (completing-read "bib: " candidates) candidates)))
```

minad:
Next step: No Embark :)

mtreca:
@minad Agreed. Could you explain a bit further how I can shortcut the embark dependency by "providing the keymap and the actions as commands. " ?

minad:
Okay, I can take another look soon. This PR is not up-to-date yet, right? Can you please push these Marginalia changes?

mtreca:
No I did not push the changes yet, I am struggling a little incorporating these changes with the interactive lambda function using `complete-with-action` (my elisp skills are not great). I will push as-is for now.

embark-bibtex.el
```diff
@@ -0,0 +1,140 @@
...
+   (bibtex-completion-candidates)))
+
+;; TODO Broken, use (cdr (assoc (completing-read "bib" candidates) candidates)) form
+;;;###autoload
+(defun embark-bibtex (bib-entry)
+  "Search BibTeX entries using `completing-read' and embark
+actions."
+  (interactive
+   (list
+    (completing-read
```
minad, line 109: This completing-read function should be factored out into a separate function `embark-bibtex--read` for example.

embark-bibtex.el
```diff
@@ -0,0 +1,140 @@
...
+  "The default action for the `embark-bibtex` command."
+  :group 'bibtex-completion
+  :type 'function)
+
+(defmacro embark-bibtex-embarkify-action (action name)
+  "Wraps the function ACTION in another function named NAME which
+extracts the key from the candidate selected in embark and
+passes it to ACTION."
+  `(defun ,name (cand)
+     (interactive "s ")
```
minad, line 79: Instead of `"s "`, use `(list (embark-bibtex--read "Some prompt: "))` here!

embark-bibtex.el
```diff
@@ -0,0 +1,140 @@
...
+   (list
+    (completing-read
+     "BibTeX entries: "
+     (lambda (string predicate action)
+       (if (eq action 'metadata)
+           '(metadata (category . bibtex))
+         (complete-with-action action (embark-bibtex--get-candidates) string predicate))))))
+  (apply embark-bibtex-default-action (list bib-entry)))
+
+(embark-define-keymap embark-bibtex-map
```
minad, line 117: Don't use embark-define-keymap here, just manually define a keymap via make-sparse-keymap and so on. It won't take much more code.

embark-bibtex.el
```diff
@@ -0,0 +1,140 @@
...
+  ("c" embark-bibtex-insert-citation)
+  ("r" embark-bibtex-insert-reference)
+  ("k" embark-bibtex-insert-key)
+  ("b" embark-bibtex-insert-bibtex)
+  ("a" embark-bibtex-add-PDF-attachment)
+  ("e" embark-bibtex-edit-notes)
+  ("s" embark-bibtex-show-entry)
+  ("l" embark-bibtex-add-pdf-to-library))
+
+(add-to-list 'embark-keymap-alist '(bib . embark-bibtex-map))
```
minad, line 130: 4. At this point, this line is the *only dependency* with Embark. Remove it.  Then you are completely independent of Embark. This line can be written in the user config in order to make the keymap available to embark.

bdarcus: So in the end, the idea for a user who has selectrum setup as their completion framework, they just install this package (or, ideally at some point, just bibtex-completion), and install and configure embark to use this keybinding, as so associated the bibtex-completion commands with the selected candidates?

mtreca:
@minad I added your remarks to a new version. My issue is that embark actions do not seem to be recognized anymore...

bdarcus:
I just tested this; just initially loading the bib and narrowing. 

One little thing: the matching is ordered? E.g. author, then title, etc.

And if there's no author, I actually can't sort on the rest (like title).

Where should this best be addressed, and how? In the users selectrum setup? In this package?

I was just testing this with emacs-sandbox, and the most minimal configuration.

mtreca:
@bdarcus I am pretty sure candidate ordering is outside of the scope of this project. I personally use [prescient](https://github.com/raxod502/prescient.el) but the [orderless](https://github.com/oantolin/orderless) package is also popular.

Did you try a fuzzy completion-style?

bdarcus:
> @bdarcus I am pretty sure candidate ordering is outside of the scope of this project. I personally use [prescient](https://github.com/raxod502/prescient.el) but the [orderless](https://github.com/oantolin/orderless) package is also popular.

Yes, you are right; if I load selectrum-prescient, it works as I expect!

mtreca:
> Edit: when the key-map is working, how does one invoke it from the selected candidate?

That's a very good question, I am not really sure. By using embark, you can define a shortcut key to trigger `embark-act` like so:

```elisp
(use-package embark
  :ensure t
  :bind (:map selectrum-minibuffer-map
         ("M-o" . embark-act)))
```

You also have to manually add the map to the embark keymap as I removed the explicit dependency:

```elisp
(add-to-list 'embark-keymap-alist '(bib . embark-bibtex-map))
```

(Please note that the keymap is not picked-up by embark-actions yet, this probably has to do with the recent refactoring).

That being said, I don't think there is a built-in way to call different actions on candidates in Emacs. I guess the long-term plan is to have multiple action and completions available to users (one per action), in which case embark would not be required since one could simply call `bibtex-completion-open-pdf` and use standard completion there without needing to select an action.

minad:
@mtreca Yes, this is right. You have to call the action via embark-act! There is no built-in way to call different actions on candidates without Embark, but the embarked actions are actually proper commands now, so you can call them directly via M-x as you already wrote with bibtex-completion-open-pdf. It is not as comfortable but it is the Emacs default way. I am not claiming that Emacs is comfortable by default, for me it is barely usable honestly.

embark-bibtex.el
```diff
@@ -76,7 +88,8 @@
 extracts the key from the candidate selected in embark and
 passes it to ACTION."
   `(defun ,name (cand)
-     (interactive "s ")
+     (interactive
+      (list (embark-bibtex--read "BibTeX entries: ")))
```
minad, line 92: Maybe you want to configure a different prompt here per embarkified command?

mtreca: `embark-bibtex--read` should not take a prompt argument at all, no?

mtreca:
So, I changed a few things in between iterations.

- Since embark is no entirely optional, I renamed the module and function prefixes to `bibtex-actions`
- The macro now autonames functions based on their base names, and take docstrings (I still have to document those).
- I fixed the action selection for those functions, which should now work properly.

I still don't see why embark actions do not show up anymore when calling `bibtex-actions`, but I will figure this out later.

embark-bibtex.el
minad: Oh, it should. For each action a separate prompt. This ensures that the commands are all usable in standalone moe.

minad:
Great work! :+1: 

* Since embark is no entirely optional, I renamed the module and function prefixes to bibtex-actions

My suggestion would be bibtex-completion. But everything should be working first before making this proposal again :laughing:

bdarcus:
> My suggestion would be bibtex-completion

I was thinking the same thing!

@mtreca - let me know when you've sorted out the embark-act detail, and I can test the actions.

minad:
Well, did you use *bibtex* here?

~~~ elisp
(add-to-list 'embark-keymap-alist '(bibtex . embark-bibtex-map))
~~~

In the comment above by @mtreca he used *bib*, but the completion category specified in the completing-read call is *bibtex*.

bdarcus:
Thanks @minad.

This minimal code, with emacs-sandbox, does now give me access to the commands.

``` elisp
(load "/tmp/bib/bibtex-actions.el")

(setq bibtex-completion-bibliography
      '("~/org/bib/academic.bib"
        "~/org/bib/data.bib"))

(use-package embark
  :ensure t
  :bind
  ("M-o" . embark-act))

(add-to-list 'embark-keymap-alist '(bibtex . bibtex-actions-map))
```

But embark doesn't present me with a menu of those actions, which is what I was expecting, a la ivy-bibtex.

Am I missing some config detail on embark, is something missing from this package, or just not supported?

minad:
Did you setup Embark properly with which-key? Embark uses which-key as menu. But you can also use a completing-read prompter. Please take a look at the Embark README.

minad:
Just FYI for the full setup I am using Selectrum(Icomplete-vertical) + Marginalia + Embark + Orderless(Prescient) + Which-key + Consult. In parens the alternatives. I am not sure if I can point you to a unified resource for how to setup all these since the packages are still changing. But if you copy the basic configurations from the READMEs of these projects or should work.

minad:
For quick experiments regarding this PR just use emacs -Q + Embark(for actions) + Which-Key(for visual menu). And maybe set the completion styles to substring.

bdarcus:
> Did you setup Embark properly with which-key?

No, but I just did, and that now works. Thank you.

It would probably be worth adding some of these details to the comments, once this gets farther along.

mtreca:
@minad Good catch regarding the use of 'bib' and 'bibtex', this is why my embark actions where not available.

I tested the new iteration of this PR and everything seems to work on my side. So there are now two entry points to this front-end:

- `bibtex-actions` (the name should be changed), which functions similarly to `ivy-bibtex` and `helm-bibtex` provided embark is enabled for bibtex actions.
- Individual bibtex actions, which call `completing-read` on a specific `bibtex-completion` function.

I guess the second group of actions could indeed easily be added to the `bibtex-completion` core.

One corner case that I still need to figure out is to make sure that `(bibtex-completion-init)` is called before any `bibtex-actions` funciton is called. Should be fixed today.

bibtex-actions.el
```diff
@@ -0,0 +1,126 @@
...
+  "Read bibtex-completion entries for completion."
+  (completing-read
+   "BibTeX entries: "
+   (lambda (string predicate action)
+     (if (eq action 'metadata)
+         '(metadata (category . bibtex))
+       (complete-with-action action (bibtex-actions--get-candidates) string predicate)))))
+
+;;;###autoload
+(defun bibtex-actions (bib-entry)
```
bdarcus, line 110:
```suggestion
(defun bibtex-actions (local-bib)
```
The parameter here is still misleading ;-)

minad:
@mtreca 

> bibtex-actions (the name should be changed), which functions similarly to ivy-bibtex and helm-bibtex provided embark is enabled for bibtex actions.
> Individual bibtex actions, which call completing-read on a specific bibtex-completion function.
> I guess the second group of actions could indeed easily be added to the bibtex-completion core.

I wonder a bit about this design, what makes the default-action now different from the individual actions. For me this is a flat list of commands. There is bibtex-completion-insert, bibtex-completion-open-pdf, ...
This means the user just binds the one or two actions they use the most to some keys and the other actions are always available. There is no real advantage in going through the single action entry point. Actually it is rather the advantage of this design that you have the option to bind multiple commands directly for faster access.

> One corner case that I still need to figure out is to make sure that (bibtex-completion-init) is called before any bibtex-actions funciton is called. Should be fixed today.

Call it in bibtex-action-read? Is this similar to bookmark-maybe-load, which loads the bookmark list?

bdarcus:
> There is no real advantage in going through the single action entry point. Actually it is rather the advantage of this design that you have the option to bind multiple commands directly for faster access.

In the end, I guess it's just a UX question; what's most intuitive and convenient for the user?

The single entry point is basically the user saying "I want to search my bib collection."

The specific commands are them saying "I want to insert a citation key/open a PDF/etc."

The specific commands wouldn't be changing with the removal of the single entry point; right?

I'm agnostic/not sure.

minad:
> The specific commands wouldn't be changing with the removal of the single entry point; right?

They would not change. All I am saying is the following: I have a list of commands operating on bibtex entries. I will bind the most commonly used command to a specific key (the default action by my own choosing, maybe insert) and then I have another often used action (maybe open pdf), which I bind to another key. All the other actions can be accessed via the Embark keymap or via M-x. I just don't see why there is a special default action on the level of bibtex-completion and in this PR. This makes sense if you have the action/command distinction - in Helm and Ivy it makes sense, in this PR not so much I think.

For this reason I would propose to remove the `bibtex-actions-default-action` default action. The user can just bind the action they want to some global keymap, e.g., `C-c b`. There is no reason for an additional defcustom. Furthermore `bibtex-actions` can be removed. All I am proposing here is a simplification without losing any functionality.

minad:
The next logical steps could be the following:

1. Add the interactive spec directly to all of the bibtex-completion actions.

~~~ elisp
(interactive (list (cdr (assoc (bibtex-actions--read) (bibtex-actions--get-candidates)))))
~~~

I would maybe clean this up a bit, such that it is just  `(interactive (list (bibtex--completions-read)))`. The actions don't have an interactive spec as of now, so this is a pure addition.

2. The keymap could also go to bibtex-completions. This is justifiable. You can bind this keymap as a prefix map to `C-c b`. Then `C-c b p` would open pdf and so on. This would also make sense even if you don't use Embark, just plain Emacs. 

In the end one would not even need a separate package. This was my original proposal :)

bdarcus:
I'm convinced!

So then making this available for selectrum users wouldn't require any code; just a short section of the bibtex-completion docs.

minad:
Let's see what @mtreca and @tmalsburg think about it. 

Maybe there is one thing which would be required to make this "fair" to helm and ivy users. The `bibtex--completions-read` function which goes into the interactive spec, should probably be written such that you can configure the completion backend. The default implementation based on completing-read from this PR should go the bibtex-completion package and the helm and ivy packages could provide their respective versions. This way you would have all bibtex commands available via M-x in all completion systems and all bibtex commands could also be triggered as actions in each completion system. You would get the same power with each of the completion systems.

mtreca:
I do agree with @minad regarding the removal of the defcustom which does not seem to serve a purpose anymore.

However, I think that @bdarcus makes a valid point regarding the design decision of removing a single entry point. Suppose I have this configuration, related the workflow mentioned by @minad, which I think a lot of users would be following:

```elisp
(use-package bibtex-completion
  :bind (("C-c b c" . bibtex-actions-insert-key)
         ("C-c b p" . bibtex-actions-open-pdf)))
```

In the rare occasions where I would need to call a `bibtex-actions` command that is not bound, do you think it makes more sense to call it from one of the bound functions and toggle the embark actions, or to use a specific command? I do agree that this distinction is artificial since the embark keymap is accessible from any `bibtex-actions` command, but it might be counter-intuitive to some to call, for instance, `bibtex-actions-open-pdf` to actually call another command through embark that has nothing to do with PDFs.
This could be for instance confusing for users switching from ivy or helm (such as myself), who are expecting a generic `ivy-bibtex` function to trigger actions on.

That being said, this functions would really be a placeholder (no default actions, just a completion view with embark support), and removing it would simplify the code further. I would also like to hear what @tmalsburg thinks of this.

minad:
> In the rare occasions where I would need to call a bibtex-actions command that is not bound, do you think it makes more sense to call it from one of the bound functions and toggle the embark actions, or to use a specific command?

In this case I would probably call `M-x bibtex-the-command-I-want` or I would know that I can call any bibtex command and just take a look at the Embark actions.

> This could be for instance confusing for users switching from ivy or helm (such as myself), who are expecting a generic ivy-bibtex function to trigger actions on.

I don't think it would be confusing as soon as you get accustomed to the Embark work flow, which is basically just the standard Emacs workflow. Everything is just a command. Embark only offers additional quick access commands on top via the action menu. You can always execute exactly the command you want via M-x.

> That being said, this functions would really be a placeholder (no default actions, just a completion view with embark support), and removing it would simplify the code further. 

Yes, it is only a no-op placeholder. And theThat being said, this functions would really be a placeholder (no default actions, just a completion view with embark support), and removing it would simplify the code further. 

> That being said, this functions would really be a placeholder (no default actions, just a completion view with embark support), and removing it would simplify the code further. 

Yes, it is only a no-op placeholder or it would execute a "default action". But then it makes more sense to directly invoke the default action command.

Regarding the confusion - one point I would like to add - I suggest it is better to make the package somehow adapt to the Embark/Emacs way instead of trying to implement its own paradigm. This is what a more experienced user of the package suite would expect I guess. If you are looking at it from the perspective of the helm-bibtex package, it may be confusing, but I don't think it would be in the greater scheme of things. People transitioning from Ivy/Helm to Embark already have some adaption to make, and under such circumstances I would value consistency. I think the main thing which has to click is that Embark is really only normal commands, I mean it is super simple. Note that you can even call `M-o M-x some-command`, if M-o triggers Embark in your case - redirecting to any different action command you like. The keymaps are actually only a convenience feature :)

bdarcus:
> 1. Add the interactive spec directly to all of the bibtex-completion actions...  
> I would maybe clean this up a bit, such that it is just `(interactive (list (bibtex--completions-read)))`.

So to boil this down, and to tie back to your [earlier suggestion](https://github.com/tmalsburg/helm-bibtex/issues/353#issuecomment-767602801), and [this more recent one](https://github.com/tmalsburg/helm-bibtex/pull/355#issuecomment-774512404), we want this, where `bibtex--completions-read` then uses ivy or helm in their respective packages, but defaults to `completing-read` in bibtex-completion?

``` elisp
(defun bibtex-completion-open-any (keys)
  (interactive (list (bibtex--completions-read))
  (bibtex-completion-open-pdf keys 'bibtex-completion-open-url-or-doi))
```

mtreca:
> > One corner case that I still need to figure out is to make sure that (bibtex-completion-init) is called before any bibtex-actions funciton is called. Should be fixed today.

> Call it in bibtex-action-read? Is this similar to bookmark-maybe-load, which loads the bookmark list?

I does look similar. This is what I ended up doing.

> I would maybe clean this up a bit, such that it is just (interactive (list (bibtex--completions-read))). The actions don't have an interactive spec as of now, so this is a pure addition.

Just to make sure I understand what you are suggesting: We should add a `(bibtex--completions-read)` function that, depending on the chosen backend, will provide adequate candidates to the completion frontend (`helm`, `ivy`, or standard `completing-read`). That way, pretty much all of this PR can be directly integrated into `bibtex-completions`. Am I correct?

If so, I really like this idea, it would be a pretty elegant solution since:

- There would be no need for an additional module file, and potentially another MELPA module as initially suggested.
- No external dependencies are needed, not even `embark`, as all completion commands are now directly accessible without having to have a completion action selection mechanism enabled (ivy, helm, embark).
- Potentially simpler code in to `helm-bibtex` and `ivy-bibtex` frontends.

minad:
> Just to make sure I understand what you are suggesting: We should add a (bibtex--completions-read) function that, depending on the chosen backend, will provide adequate candidates to the completion frontend (helm, ivy, or standard completing-read). That way, pretty much all of this PR can be directly integrated into bibtex-completions. Am I correct?

Yes.

> There would be no need for an additional module file, and potentially another MELPA module as initially suggested.

Yes. No need for another package.

 > No external dependencies are needed, not even embark, as all completion commands are now directly accessible without having to have a completion action selection mechanism enabled (ivy, helm, embark).

Yes, it would work out of the box with Emacs default completion. But if you install the helm-bibtex package you get helm actions, if you install the ivy-bibtex package you get ivy actions and if you install embark you can use the keymap as embark action keymap.

> Potentially simpler code in to helm-bibtex and ivy-bibtex frontends.

I am not sure about that, maybe. But very important - there should be no degradation in functionality for Helm or Ivy. Rather you can access all the bibtex commands directly instead, where all of them have all additional actions. The ui would be consistent over all the completion systems, which I find desirable.

tmalsburg:
> One little thing: the matching is ordered? E.g. author, then title, etc.

Bibtex-completion returns entries in the reverse order in which they appear in the .bib file.  So when I add a new entry at the end of the file, it appears at the top of the list.  Personally, I never felt we needed sorting capabilities and over the years there was only one FR regarding sorting (I think).  One issue with sorting is that some people have really large bibliographies (e.g. [crypto.bib](https://cryptobib.di.ens.fr/)) and sorting be too slow when done ad-hoc.

minad:
@bdarcus 

> One little thing: the matching is ordered? E.g. author, then title, etc.
> And if there's no author, I actually can't match on the rest (like title).

I think that would work against the design of some completion styles. For example there is the completion style named [orderless](https://github.com/oantolin/orderless), which is not about the order of the candidates, but rather about the order of the space-separated matching terms. You can match in arbitrary order, first write some author name, then the a part of the title or the other way round. This allows for a lot of flexibility and faster narrowing in my opinion. I actually recommend Orderless in combination with Selectrum, Embark, Consult etc. This is better than the built-in completion-styles like substring, basic, flex etc.

bdarcus:
> Sorting, to me, seems like a UI feature. So it should probably not be part of bibtex-completion

Yeah, totally fine with that. I had just needed to setup selectrum to use prescient (or, per @minad, orderless).

tmalsburg:
Interesting discussion and cool to see how this project is coming along.  I'm trying to keep up with this thread, but unfortunately I don't have time to read everything carefully.  If you need input from me specifically, please @ me.

Regarding the architecture: I'm not convinced that UI code should go into bibtex-completion.  As I understand it, the idea is to make the action-functions, which are currently pure API, into proper Emacs commands.  This seems elegant because it is minimally invasive, but I'm afraid that it will complicate maintenance of this code in the future.  For instance, if we'd like to change the UI of the completing-read front end, it would affect the API as well and hence helm-bibtex and ivy-bibtex.  Likewise, if we need to change the API, it will affect the UI of the completing-read front end.  Not ideal.  Also, users of helm/ivy-bibtex would effectively get two UIs, the help/ivy front ends plus the action commands.  Also keep in mind that there are other packages that depend on bibtex-completion as well and which might also be affected by these changes (org-ref, org-roam-bibtex).

As far as I’m concerned it’s best to keep things modular, the features orthogonal, and the code as easy to maintain as possible for everyone involved.  If that means that we'll have a couple of extra lines of code and a new package, that's an okay price to pay.

By the way, I'd call the new package completing-read-bibtex in line with helm-bibtex and ivy-bibtex.  That’s largely self-explanatory for anyone who’s made contact with helm- and ivy-bibtex.

Once again, great work, and I'm looking forward to seeing the result.

minad:
I am pretty sure it would not complicate maintenance, break compatibility and so on. And as I said the changes are minimally invasive. I am a bit disappointed that I or we seemed to have failed to convince you here. For sure it is okay to keep the current design. I believe it could be better, but it is your decision.

tmalsburg:
Perhaps, I'm misunderstanding the proposal.  But let's say it was convenient to rename a user-facing function, e.g. because users perceive it to be misleading, how could we change that function without also changing the API for all other packages using bibtex-completion?  When I created these functions they were not intended to be user-facing, and accordingly I didn't put much thought into their names.  For instance, `bibtex-completion-open-any` is not great.  I also don't see (could be my fault) the downside of placing the proposed front-end in a separate file.  The cost of that is basically zero and for users it doesn't really make a difference.

bdarcus:
Regardless of what the final decision is (separate package or not), on this point I made earlier:

> It would probably be worth adding some of these details to the comments/docs, once this gets farther along.

... the documentation should recommend prescient and/or orderless, and give an example configuration with (including setting up the keymap) and without embark?

mtreca:
Thanks for your feedback @tmalsburg.

I think keeping things separated for now is OK. We could merge a new module, `completing-read-bibtex`, containing the current implementation, and document its possible integration with `embark`. That way, `completing-read` and `selectrum` users get a working front-end to `bibtex-completion` that re-uses a lot of builtin Emacs functionality without impacting `ivy-bibtex` and `helm-bibtex` users. This is a win-win situation IMO.

(This PR is not ready to merge yet, I still need to document the new module and "port" local bib functionality from `ivy-bibtex`).

On the longer run, I do agree with @minad's proposed implementation, but it might be outside of the scope of this PR (and of my elisp abilities). Maybe another PR could be opened regarding this issue, in which case I would be interested in participating.

bdarcus:
> it might be outside of the scope of this PR (and of my elisp abilities).

I'd guess this possible new PR would include the modifications (as demonstrated on this PR) to bibtex-completion, as well as helm/ivy-bibtex (just the one new function?).

Functionally, it would be the same as this package for the end user (since @mtreca already removed the single entry point from this PR), but may have other advantages.

minad:
Fyi in a recent Consult discussion I linked above (see in particular this comment https://github.com/minad/consult/issues/210#issuecomment-775923226), Embark got support for non-command actions, since this allows to retain text-properties and makes it easier to write custom actions in certain cases. This also applies to your use case, where everything is going through the single entry point as per the Ivy/Helm-UI. I still believe that the "everything is a command" paradigm is better since it gives the user more flexibility to bind the commands to keys and it works with just Emacs built-ins, without installing any package. But there is the alternative now.

For more details - Embark allows to bind these two types of actions now:

~~~ elisp
(defun function-action (target)
  ;; non interactive, target retains text properties
   ...)

(defun command-action (target)
   (interactive (list (completing-read ...))) ;; target is a plain string WITHOUT text properties
   ...)
~~~

bdarcus:
Nice to see how quickly this ecosystem is evolving. But on this ...

> This also applies to your use case, where everything is going through the single entry point as per the Ivy/Helm-UI.

... @mtreca removed that single entry point. Not sure what he's thinking about it's status as he wraps up final revisions, but doesn't seem he needs to add it back.

mtreca:
I am quite busy with work this week, but I will definitely look at this discussion in detail when I have the chance. As @bdarcus said, nice to see the embark package evolving quickly!

bdarcus:
I was wanting to test out the UX. How do I associate the kb map with `embark-act`?

bdarcus:
I was just testing the UX, to see how intuitive I find it. A few observations:

First, I do think `bibtex-insert-key` would be more straightforward a command than `completing-read-bibtex-insert-key`.

Screenshot from 2021-02-15 16-51-05

Second, it would be nice if the bibtex commands were grouped at the topic of this list. Is that something the user would be configuring?

Screenshot from 2021-02-15 19-04-43

Finally, a single command entry point: I don't think it's needed, particularly if the individual commands are simply and straightforward.

bdarcus:
> I'd guess this possible new PR would include the modifications (as demonstrated on this PR) to bibtex-completion, as well as helm/ivy-bibtex (just the one new function?).

I don't have the skill to do this, and so don't feel comfortable opening a new PR ATM.

But I did try to adapt the new `bibtex-completion--read` function @mtreca came up with, and then add the single line that @minad suggested to `bibtex-completion-insert-key`.

https://github.com/bdarcus/helm-bibtex/commit/81a76ac85fd3c8e401760485a19aa212680e6e1d

It almost works, but I get a "wrong type" error when selecting a candidate.

If we did want to open a new PR on this, seems the next steps would be to fix these functions, and adapt it to helm/ivy-bibtex. Would obviously need to alleviate Titus' concerns for him to reconsider.

bdarcus:
[Above](https://github.com/tmalsburg/helm-bibtex/pull/355#issuecomment-774504526), @minad suggested the following:

> You can bind this keymap as a prefix map to C-c b. Then C-c b p would open pdf and so on.

If I understand right, wouldn't that lead to a UI when invoking `embark-act` (`C-c` in his example?) where you have two options initially:

1. Hit return for the default action.
2. Hit `b` to bring up a secondary list of bibtex-completion-specific alternatives.

If yes, wouldn't this be better from a UI POV, and partially address my point above about command names?

bdarcus:
@tmalsburg - on the naming question, you suggested this:

> By the way, I'd call the new package completing-read-bibtex in line with helm-bibtex and ivy-bibtex. That’s largely self-explanatory for anyone who’s made contact with helm- and ivy-bibtex.

Wouldn't that result in/require unnecessarily verbose command names? What do you think about cr-bibtex?

bibtex-actions.el
```diff
@@ -0,0 +1,116 @@
...
+  (mapcar
+   (lambda (cand)
+     (cons (bibtex-completion-format-entry cand (1- (frame-width)))
+           (cdr (assoc "=key=" cand))))
+   (bibtex-completion-candidates)))
+
+(defun bibtex-actions--read ()
+  "Read bibtex-completion entries for completion."
+  (bibtex-completion-init)
+  (completing-read
```
bdarcus, line 32: Should this be `completing-read-multiple`?

bdarcus:
@minad - when you get a chance, a few questions:

### Should we use `completing-read-multiple`? + related `embark-collect-snapshot` question

So I asked above about `completing-read-multiple`, but let me ask the question from a broader and more user-focused perspective.

I am writing a paper. I want to work with a a collection of references, and insert them as citations in different places within the text.

Sometimes those citation are single key, and other times multi key.

```
A single key citation [doe15].
A multi key citation [doe20, smith17, smith19].
```

What's the preferred way to handle this here for the multi key example?

1. select multiple candidates using `completing-read-multiple`, run insert-key action on them
2. use `embark-collect-snapshot`, run the default action on the ones I need (though see below for a caveat)
3. allow both options?

My thought is we want 3, but ...

#### embark-collection-snapshot non-default actions?

cc @oantolin

I'm confused on 2. The embark documentation says the following:

> The embark-collect-snapshot command produces a buffer listing all the current candidates, for you to peruse and run actions on at your leisure. 

This suggests to me that I can access any action from the snapshot, which is what we want. 

But while I can access the default action, I can't figure out how to access others actions (say open the PDF) from that view, which would be ideal for this use case. 

Seems somehow that documentation could use amending to clarify. And if one can only access the default action, is this restriction necessary?

If we should use `completing-read-multiple` instead, how do we deal with the UI problem I note [above](https://github.com/tmalsburg/helm-bibtex/pull/355#issuecomment-779503691) (see the first screenshot)?

And if we don't use `completing-read-multiple` and a user isn't using embark, then they would have to run multiple cycles of command-complete-select action, which seems less-than-ideal.

Seems to me we want 3, ideally with embark allow different actions on the snapshot items, but I'd appreciate your thoughts.

### prefix map submenus in embark

The other issue I noted [above](https://github.com/tmalsburg/helm-bibtex/pull/355#issuecomment-782304261) also. I can't get the prefix map submenu to display.  Is that possibly related to this bug your reported?

https://github.com/oantolin/embark/issues/139

How should we configure the prefix map so it works as intended?

oantolin:
Let me start with the caveat that I have only skimmed the conversation here, and haven't read it the whole thing carefully, nor have I looked at the code yet.

> This suggests to me that I can access any action from the snapshot, which is what we want.

> But while I can access the default action, I can't figure out how to access others actions (say open the PDF) from that view, which would be ideal for this use case. The embark-act keybinding, for example, doesn't work.

> Seems somehow that documentation could use amending to clarify. And if one can only access the default action, is this restriction necessary?

You can definitely use all the actions you have from an Embark collect buffer (indeed, the buffers wouldn't be very useful if you couldn't!). By default in the Embark collect buffers `embark-act` is bound to `a`, but if you have a global keybinding for `embark-act` you should be able to use that too. And of course, there is always `M-x embark-act` too.

I don't know what is keeping this from working for you, but we'll figure it out. So what exactly happens when you try to run `embark-act` in the Embark collect buffer? Also, does this inability to use actions in the Embark collect buffer only happen with your bibtex actions or is it a general problem with your installation? Have you tried `embark-collect-snapshot` from, say `find-file` and then tried to run some actions from the collect buffer?

> The other issue I noted above also. I can't get the prefix map submenu to display. Is that possibly related to this bug your reported?

I think this does sounds like exactly the issue that @minad reported. I know how to fix it but haven't done it yet since it is a fairly large change. (I hadn't even noticed the problem because I don't use which-key myself.) Knowing that more people are using prefix keymaps in Embark, I'll make more of an effort to find the time.

However, I don't understand why you are using a prefix keymap here. Am I understanding correctly that you are placing every single bibtex action behind the prefix `b`? So no matter what action a user of your package wants to use, they press whatever keybinding they have for `embark-act`, then `b`, then finally select the action? If that's the case I highly recommend skipping the `b`, it sounds unnecessary and inconvenient.

It looks like you already have all the actions in a keymap called `bibtex-actions-map`. Then to use with Embark, all a user of your package needs to do to access the actions without a prefix key is:

```elisp
(setf (alist-get 'bibtex embark-keymap-alist) 'bibtex-actions-map)
```

bdarcus:
> Let me start with the caveat that I have only skimmed the conversation here, and haven't read it the whole thing carefully, nor have I looked at the code yet.

Yeah, you shouldn't need to read the conversation beyond that one message. Thanks!

> So what exactly happens when you try to run `embark-act` in the Embark collect buffer? Also, does this inability to use actions in the Embark collect buffer only happen with your bibtex actions or is it a general problem with your installation? Have you tried `embark-collect-snapshot` from, say `find-file` and then tried to run some actions from the collect buffer?

Context: using Doom evil, and the WIP selectrum et al module. 

https://github.com/hlissner/doom-emacs/pull/4664/

If I hit "a" within the snapshot buffer on `find-file`, the cursor just changes to insert mode.

If I do the `embark-act` keybinding instead, it opens a previously opened buffer.

That's basically what I see when using the bibtex commands also.

> > The other issue I noted above also. I can't get the prefix map submenu to display. Is that possibly related to this bug your reported?
> 
> I think this does sounds like exactly the issue that @minad reported. I know how to fix it but haven't done it yet since it is a fairly large change. (I hadn't even noticed the problem because I don't use which-key myself.) Knowing that more people are using prefix keymaps in Embark, I'll make more of an effort to find the time.

No problem at all. I'm just glad we know the problem, and it wasn't me!
 
> However, I don't understand why you are using a prefix keymap here. Am I understanding correctly that you are placing every single bibtex action behind the prefix `b`? So no matter what action a user of your package wants to use, they press whatever keybinding they have for `embark-act`, then `b`, then finally select the action? If that's the case I highly recommend skipping the `b`, it sounds unnecessary and inconvenient.

OK, we'll do that.

The issue I was trying to work around I note above in the "sorting/grouping" section of [this post](https://github.com/tmalsburg/helm-bibtex/pull/355#issuecomment-779503691). Basically, too-long command names (which is related to an orthogonal question about package name), and wanting to sort or group to keybindings so easier to quickly grok.
 
> It looks like you already have all the actions in a keymap called `bibtex-actions-map`. Then to use with Embark, all a user of your package needs to do to access the actions without a prefix key is:
> 
> ```emacs-lisp
> (setf (alist-get 'bibtex embark-keymap-alist) 'bibtex-actions-map)
> ```

oantolin:
So basically the Embark collect buffers don't let you use other actions at all under your Doom Emacs configuration, @bdarcus? Since I don't use Doom, maybe you can help me debug this issue. First, just to check the obvious possible problem: you are sure you are running `embark-act`, right? There isn't some other Doom keybinding shadowing your keybinding for `embark-act`? Did you try running it by name instead of by keybinding? (I don't know if `M-x` works in Doom, but if it doesn't, maybe `:` runs arbitrary Emacs commands too?)

bdarcus:
> Did you try running it by name instead of by keybinding?

Ugh; I should have thought of that!

If I run the command directly, it works fine. I'll report it to that doom PR.

Thanks again @oantolin!

oantolin:
That's a relief, @bdarcus! It's sounds like it's just a keybinding or Evil state issue, then, which should probably just be fixed locally in your personal configuration. You should bind some key to run `embark-act` in the Evil normal state in buffers that are in `embark-collect-mode`, or alternatively you could configure things so that `embark-collect-mode` buffers start in Evil's Emacs state (but you'd loose your nice Evil navigation, so probably the first choice is better).

bdarcus:
Doom has extensive keybinding support, including in that module for embark-act, so it should really be fixed there. I'll link your comment over there.

oantolin:
Oh, I went to hlissner/doom-emacs#4664. It also loads Embark! So then the Evil normal state for embark-act in Embark collect buffers should be added there, not in your personal configuration.

bdarcus:
So for the broader questions I was asking, the simplest approach is to keep things as are with just `completing-read`, and encourage people to use `embark` and `embark-collect-snapshot` if they need more flexibility to act on multiple candidates quickly and easily.

Could always add `completing-read-multiple` support later if demand for it?

minad:
It seems @oantolin already answered most questions satisfactorily? Thank you, @oantolin! 

@bdarcus Regarding `completing-read-multiple` - I believe there is no special support as of now. Therefore, as you said, everything should be kept as is.

Maybe Embark could get support for `completing-read-multiple` such that one could act on the already selected candidates at once. Independent of `completing-read-multiple` one could add an Embark feature which allows selecting candidates and then act on all the selected candidates. This could work with plain `completing-read`. I am unsure if this has already been considered - it would be quite a deviation from the current approach.

oantolin:
@bdarcus About using the `b` prefix for all actions you wrote:

> The issue I was trying to work around I note above in the "sorting/grouping" section of [this post](https://github.com/tmalsburg/helm-bibtex/pull/355#issuecomment-779503691). Basically, too-long command names (which is related to an orthogonal question about package name), and wanting to sort or group the keybindings so easier to quickly grok.

I think probably a better fix would be to hide all the general actions (that is, actions bound in `embark-general-map`) from the which-key display. There are so many now! And each of them is useful, I believe, so I like having them there, always available, but since they are always there, maybe users don't need to be constantly reminded of them.

bdarcus:
> I think probably a better fix would be to hide all the general actions (that is, actions bound in `embark-general-map`) from the which-key display.

At first glance, I like that idea @oantolin. 

Would we suggest that here, say in an example configuration in the docs?

bibtex-actions.el
```diff
@@ -0,0 +1,116 @@
+;;; bibtex-actions.el --- Bibliography manager action
+
```
bdarcus, line 2:
```suggestion
;;; Commentary:
;;;
;;; Provides completing-read-based interactive command wrappers around
;;; bibtex-completion commands, to make these actions available as standard emacs
;;; commands, and within completing-read-compatible completion system.
```

bibtex-actions.el
```diff
@@ -0,0 +1,116 @@
+;;; bibtex-actions.el --- Bibliography manager action
```
bdarcus, line 1:
```suggestion
;;; bibtex-actions.el --- Bibliography manager actions
```

bdarcus:
> Maybe Embark could get support for `completing-read-multiple` such that one could act on the already selected candidates at once. Independent of `completing-read-multiple` one could add an Embark feature which allows selecting candidates and then act on all the selected candidates.

For this example, I also think the selectrum UI for multiple selection would need work. It would seem to be better, at least optionally, to able to mark the selected candidates in the selectrum minibuffer, as in helm.

But with `embark-collect`, this is less important.

bdarcus:
> Maybe another PR could be opened regarding this issue, in which we could maybe convince you further! In any case I would be interested in participating in this other PR as well.

At @mtreca's request, I've opened (the start of) just such a PR at #361!

tmalsburg:
Hi all, I'm currently super busy at work and couldn't follow the recent discussion.  I will soon catch up and look at the PR but it may take a couple of days.

mtreca:
Same as @tmalsburg, I am really busy this month. Will take time to review these discussions and apply suggested changes.

bdarcus:
Sounds good @mtreca @tmalsburg.

To pare down the content here, I've hidden some of the recent longer posts of mine. I think this post summarizes my view of the current status.

https://github.com/tmalsburg/helm-bibtex/pull/355#issuecomment-784251233

bibtex-actions.el
```diff
@@ -0,0 +1,116 @@
...
+    (define-key map (kbd "e") 'bibtex-actions-edit-notes)
+    (define-key map (kbd "s") 'bibtex-actions-show-entry)
+    (define-key map (kbd "l") 'bibtex-actions-add-pdf-to-library)
+    map))
+
+(defun bibtex-actions--get-candidates ()
+  "Return all keys from bibtex-completion-candidates."
+  (mapcar
+   (lambda (cand)
+     (cons (bibtex-completion-format-entry cand (1- (frame-width)))
```
bdarcus, line 25: We need a better solution for this, as we're missing data in the process.

If we can't complete against the output of `bibtex-completion-candidates` using `completing-read`, can we transform it so we don't lose any of that data?

The use case here is I really want to add tags to complete against, so I can distinguish among different projects, etc.

See #362.

bdarcus: I posted the [question on reddit](https://www.reddit.com/r/emacs/comments/lsxjfx/completingread_displaying_using_one_string_but/), and got a response from one of the selectrum devs.

Seems we want to use propertize something like this.

``` elisp
(completing-read ";) "
                 (list (concat "vim" (propertize "there is evil-mode" 'invisible t))
                       (concat "emacs"  (propertize "the best" 'invisible t))))
```

So in other words, if `bibtex-completion-additional-search-fields` is set, add those fields from `bibtex-completion-candidates`, but hide them.

I guess this is roughly similar to the ivy display transformer.

bibtex-actions.el
```diff
@@ -0,0 +1,116 @@
...
+           (cdr (assoc "=key=" cand))))
+   (bibtex-completion-candidates)))
```
bdarcus, line 27: This suggestion uses `cl-loop`, as elsewhere in the codebase, and for better performance, and also uses propertize to allow us to support `bibtex-completion-additiona-search-fields`.

```suggestion
(defun bibtex-completion--get-candidates ()
  "Return all keys from bibtex-completion-candidates."
  (cl-loop
   for candidate in (bibtex-completion-candidates)
   collect
   (cons
    ;; Here use one string for display, and the other for search.
    (propertize
     (car candidate) 'display (bibtex-completion-format-entry candidate (1- (frame-width))))
    (cdr (assoc "=key=" candidate)))))
```

bdarcus:
> I think probably a better fix would be to hide all the general actions (that is, actions bound in `embark-general-map`) from the which-key display.

@oantolin - maybe fodder for your wiki: how do I do this?

PR #360, opened 2021-02-23
Better Support to Calibre bib

yuchen-lea:
There're some special characters in calibre bib:

1. multiple files are separated by `,` instead of `;` in `file` field
2. file path often contains multiple spaces

So I make some minor changes on  `bibtex-completion-find-pdf-in-field` and `bibtex-completion-get-value` to work better with bib files exported by calibre


tmalsburg:
Thank you.

tmalsburg:
Just for the record.  This PR creates problems for people whose filenames have commas.  See #371.

PR #361, opened 2021-02-23
Make bibtex-completion (more) interactive

bdarcus:
### Introduction

Over the last year, there has been a lot of activity on packages that are built around the emacs `completing-read` API. Designed for compatibility and modularity, but in collaboration, these tools together provide users flexibility and excellent functionality. 

This PR provides an alternative approach to (really just adapts code from) #355 as a means to offer bibtex-completion functionality to emacs vertically-oriented completion systems using `completing-read`, such as `icomplete-vertical` and `selectrum` (see #353).

The changes are minor, adding only about 70 very simple LOC. It follows @minad's advice to make the core functions interactive, using a new `bibtex-completion--read` function, which allows for configurable completion systems. 

By default, this function uses `completing-read`, which is already used in `bibtex-completion` in a few places.

But the intention is that one could use another function there, as the default function is bound to `bibtex-completion-read-backend` via `fset`.

So `ivy-bibtex` and `helm-bibtex` could, if there was value in this, instead use functions based on `ivy-read` and `helm-read` respectively (though this appears to be unnecessary, as they seem to work fine as is). One could even write a different `completing-read` function, if one wanted to modify the behavior locally, or in a separate package.

Also adds a keymap; `bibtex-completion-map`.

I do think this demonstrates how small a footprint this sort of change would introduce, while providing more flexibility for users and developers, but without impacting the packages that depend on it.

### Status

Everything seems to work correctly.

The basic `bibtex-completion--read` function works, and all of the action functions included in the keymap now use it. 

Here's a screenshot showing the most basic, default, installation, with `icomplete-vertical`:

Screenshot from 2021-02-28 11-21-50

With this setup, you only have access to the default command; there are no other actions/commands available.

Adding embark, however, adds that support, and so is strongly recommended. Binding the keymap to `embark-act` like so ...

``` elisp
(setf (alist-get 'bibtex embark-keymap-alist) 'bibtex-completion-map)
```

... will get you quick and easy access to additional actions, as well as the ability to snapshot sub collections and act on those as with `ivy-occur`. 

A full set of `completing-read` compatible packages to get, IMHO, the best experience using this includes the following:

- [selectrum](https://github.com/raxod502/selectrum) (narrowing and completion)
- [embark](https://github.com/oantolin/embark/) (actions)
- [consult](https://github.com/minad/consult) (commands)
- [prescient](https://github.com/raxod502/prescient.el) and [orderless](https://github.com/oantolin/orderless) (sorting and filtering)
- [marginalia](https://github.com/minad/marginalia) (minibuffer annotations)

Here is with the above installed and configured (though consult is not strictly necessary).

Screenshot from 2021-03-01 16-20-49

### Next Steps?

I could use help on the following (which will be valuable even if we go with #355 instead):

1. Reviewing the code.
2. Testing, in general, but also to see in particular that I am correct this doesn't impact `ivy-bibtex` and `helm-bibtex`
3. I'd also like to come up with a good example configuration, both with `embark` (looking into [embark-direct-action-mode, tablist](https://github.com/oantolin/embark/issues/166#issuecomment-784534410), and [hiding the general embark commands](https://github.com/tmalsburg/helm-bibtex/pull/355#issuecomment-784266618)) and without.
4. I'm wondering how to filter on items that have notes or pdfs . See #363.
5. Would we benefit from adding an annotation function (see [this comment](https://github.com/minad/marginalia/issues/58#issuecomment-782180044))? For now, I have a placeholder comment in the code. If yes, that might suggest a slightly different UI for completing-read; say putting some of the metadata in annotation, and having the candidate display itself narrower than the full frame width?

Finally, while I think out-of-scope for this PR:

6. Evaluating whether it could be helpful to have `helm-bibtex` and `ivy-bibtex` use an adapted `bibtex-completion--read`, and if so, implementing that. As I say above, the PR so far does not impact the functionality of the `helm-bibtex` and `ivy-bibtex` actions. The `bibtex-completion` commands also work, at least within `helm` (I haven't tested with`ivy`). 

### Testing

If you want to test this, but don't have the selectrum et al packages installed, a quick and simple option I recommend using `emacs -Q` is one of the selectrum [test scripts](https://github.com/raxod502/selectrum#contributor-guide), preferably the selectrum-prescient one.

All you need to do, then, is eval this code:

``` elisp
(require 'embark)
(load "~/Code/helm-bibtex/bibtex-completion.el")
(setf (alist-get 'bibtex embark-keymap-alist) 'bibtex-completion-map)
(setq bibtex-completion-bibliography "~/org/bib/academic.bib")
```
Upon running one of the commands, and narrowing the candidates, you can then run `embark-act` (in the test scripts, it is bound to `M-o`).

### Concerns

@tmalsburg has raised some reasonable concerns, which we can summarize as:

1. impact on code maintainability (what this PR is intending to test/demonstrate)
2. what to do about command names that weren't designed to be user facing?

https://github.com/tmalsburg/helm-bibtex/pull/355#issuecomment-775103918
https://github.com/tmalsburg/helm-bibtex/pull/355#issuecomment-775138438

Issue 1 I can't really assess, but is obviously the more serious issue. Hopefully this PR addresses this concern.

On 2, the function names are mostly OK, with two obvious issues: 

1. We have one function with "PDF" while all others are all lowercase. I addressed that with https://github.com/tmalsburg/helm-bibtex/pull/361/commits/a888ea50b78b3c789d94d0e04412be9c81657e47. 
2. `bibtex-completion-open-any`; I actually think it's fine, and more importantly, have a hard time coming up with something better (maybe omit the "any"?). But if we do, we could use the same solution as 1?

I suppose if we wanted to be aggressive, we could also think about:

- `bibtex-completion-open-any` ->  `bibtex-completion-open`
- `bibtex-completion-open-url-or-doi` ->  `bibtex-completion-open-link`
- `bibtex-completion-open-pdf` (per change noted above)

And for that matter consolidating around "insert" and "open" as the key action verbs.

Not sure, but it seems like something like `make-obsolete` could work?

``` elisp
;; we want new-name interactive, but `old-name` NOT interactive and deprecated
(defun old-name ()
  (message "old name"))

(defun new-name ()
  (interactive)
  (message "new name"))

(make-obsolete 'old-name 'new-name "2.0")

;; old-name function call still works
(old-name)
```

### Relation to #355

These would be functionally the same; just different command name prefixes.

@mtreca started #355, but this code is further-developed and feature-complete. I have merged the two code bases [here](https://github.com/bdarcus/bibtex-actions) as an experiment.

### Decisions

After working on this, I have concluded this should come down to a choice between two options:

1. merge this PR here
2. we develop a [separate project ](https://github.com/bdarcus/bibtex-actions/issues/12)(I don't see any benefit to anyone to host it here)

Each option is workable, I think, and each has trade-offs.

I suppose there's also a third option, which is merging this into an experimental branch for longer term consideration, and also doing the separate package now.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -68,7 +68,7 @@
     (define-key map (kbd "r") 'bibtex-completion-insert-reference)
     (define-key map (kbd "k") 'bibtex-completion-insert-key)
     (define-key map (kbd "b") 'bibtex-completion-insert-bibtex)
-    (define-key map (kbd "a") 'bibtex-completion-add-PDF-attachment)
+    (define-key map (kbd "a") 'bibtex-completion-add-pdf-attachment)
     (define-key map (kbd "e") 'bibtex-completion-edit-notes)
     (define-key map (kbd "s") 'bibtex-completion-show-entry)
     (define-key map (kbd "l") 'bibtex-completion-add-pdf-to-library)
@@ -1379,9 +1379,15 @@ Self-contained means that cross-referenced entries are merged."
              concat
              (format "  %s = {%s},\n" name value)))))
 
-(defun bibtex-completion-add-PDF-attachment (keys)
+(defun bibtex-completion-add-pdf-attachment ()
+  ; this is basically a lowercase command alias for the function
+  ; ideally, that name should be deprecated in favor of this one
   "Attach the PDFs of the entries with the given KEYS where available."
   (interactive (list (bibtex-completion--read)))
+  (bibtex-completion-add-PDF-attachment))
+
+(defun bibtex-completion-add-PDF-attachment (keys)
+  "Attach the PDFs of the entries with the given KEYS where available."
   (dolist (key keys)
     (let ((pdf (bibtex-completion-find-pdf key bibtex-completion-find-additional-pdfs)))
       (if pdf
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -56,8 +56,23 @@
 (declare-function org-element-property "org-element")
 
 (defgroup bibtex-completion nil
-  "Helm plugin for searching entries in a BibTeX bibliography."
+  "Provides searching of entries in a BibTeX bibliography, and running actions on them."
   :group 'completion)
+
+(defvar bibtex-completion-map
+  "Keymap for bibtex-completion commands"
+  (let ((map (make-sparse-keymap)))
+    (define-key map (kbd "p") 'bibtex-completion-open-pdf)
+    (define-key map (kbd "u") 'bibtex-completion-open-url-or-doi)
+    (define-key map (kbd "c") 'bibtex-completion-insert-citation)
+    (define-key map (kbd "r") 'bibtex-completion-insert-reference)
+    (define-key map (kbd "k") 'bibtex-completion-insert-key)
+    (define-key map (kbd "b") 'bibtex-completion-insert-bibtex)
+    (define-key map (kbd "a") 'bibtex-completion-add-PDF-attachment)
+    (define-key map (kbd "e") 'bibtex-completion-edit-notes)
+    (define-key map (kbd "s") 'bibtex-completion-show-entry)
+    (define-key map (kbd "l") 'bibtex-completion-add-pdf-to-library)
+    map))
 
 (defcustom bibtex-completion-bibliography nil
   "The BibTeX file or list of BibTeX files.
@@ -477,6 +492,26 @@ for string replacement."
 
 (defvar bibtex-completion-cached-notes-keys nil
   "A cache storing notes keys obtained when the bibliography was last parsed.")
+
+(defun bibtex-completion--read ()
+  "Select BibTeX entries in completion system."
+  ;; define a completion function that defaults to completing-read, but can be overridden
+  (list (cdr (assoc (bibtex-completion--completing-read) (bibtex-completion--get-candidates)))))
+
+(defun bibtex-completion--completing-read ()
+  "Read bibtex-completion entries for completion using completing-read."
+  (bibtex-completion-init)
+  (completing-read
+   "BibTeX entries: "
+   (bibtex-completion--get-candidates)))
+
+(defun bibtex-completion--get-candidates ()
+  "Return all keys from bibtex-completion-candidates."
+  (mapcar
+   (lambda (cand)
+     (cons (bibtex-completion-format-entry cand (1- (frame-width)))
+           (cdr (assoc "=key=" cand))))
+   (bibtex-completion-candidates)))
 
 (defun bibtex-completion-candidates ()
   "Read the BibTeX files and return a list of conses, one for each entry.
@@ -910,6 +945,7 @@ governed by the variable `bibtex-completion-display-formats'."
 If multiple PDFs are found for an entry, ask for the one to open
 using `completion-read'.  If FALLBACK-ACTION is non-nil, it is
 called in case no PDF is found."
+  (interactive (list (bibtex-completion--read)))
   (dolist (key keys)
     (let ((pdf (bibtex-completion-find-pdf key bibtex-completion-find-additional-pdfs)))
       (cond
@@ -928,6 +964,7 @@ called in case no PDF is found."
 
 (defun bibtex-completion-open-url-or-doi (keys)
   "Open the URL or DOI associated with entries in KEYS in a browser."
+  (interactive (list (bibtex-completion--read)))
   (dolist (key keys)
     (let* ((entry (bibtex-completion-get-entry key))
            (url (bibtex-completion-get-value "url" entry))
@@ -1087,6 +1124,7 @@ The format depends on
   (let ((format-function
          (cdr (or (assoc major-mode bibtex-completion-format-citation-functions)
                   (assoc 'default   bibtex-completion-format-citation-functions)))))
+    (interactive (list (bibtex-completion--read)))
     (insert
      (funcall format-function keys))))
 
@@ -1096,6 +1134,7 @@ The format depends on
                 (s-word-wrap fill-column
                              (concat "\n- " (bibtex-completion-apa-format-reference it)))
                 keys)))
+    (interactive (list (bibtex-completion--read)))
     (insert "\n" (s-join "\n" refs) "\n")))
 
 (defun bibtex-completion-apa-format-reference (key)
@@ -1293,11 +1332,13 @@ Surrounding curly braces are stripped."
 
 (defun bibtex-completion-insert-key (keys)
   "Insert BibTeX KEYS at point."
+  (interactive (list (bibtex-completion--read)))
   (insert
    (funcall 'bibtex-completion-format-citation-default keys)))
 
 (defun bibtex-completion-insert-bibtex (keys)
   "Insert BibTeX entries for entries in KEYS at point."
+  (interactive (list (bibtex-completion--read)))
   (insert (s-join "\n" (--map (bibtex-completion-make-bibtex it) keys))))
 
 (defun bibtex-completion-make-bibtex (key)
@@ -1320,6 +1361,7 @@ Self-contained means that cross-referenced entries are merged."
 
 (defun bibtex-completion-add-PDF-attachment (keys)
   "Attach the PDFs of the entries with the given KEYS where available."
+  (interactive (list (bibtex-completion--read)))
   (dolist (key keys)
     (let ((pdf (bibtex-completion-find-pdf key bibtex-completion-find-additional-pdfs)))
       (if pdf
@@ -1442,10 +1484,12 @@ Creates new notes where none exist yet."
 
 (defun bibtex-completion-edit-notes (keys)
   "Open the notes associated with KEYS using `bibtex-completion-edit-notes-function'."
+  (interactive (list (bibtex-completion--read)))
   (funcall bibtex-completion-edit-notes-function keys))
 
 (defun bibtex-completion-show-entry (keys)
   "Show the first entry in KEYS in the relevant BibTeX file."
+  (interactive (list (bibtex-completion--read)))
   (catch 'break
     (dolist (bib-file (bibtex-completion-normalize-bibliography 'main))
       (let ((key (car keys))
@@ -1475,6 +1519,7 @@ Creates new notes where none exist yet."
   "Add a PDF to the library for the first entry in KEYS.
 The PDF can be added either from an open buffer, a file, or a
 URL."
+  (interactive (list (bibtex-completion--read)))
   (let* ((key (car keys))
          (source (char-to-string
                   (read-char-choice "Add pdf from [b]uffer, [f]ile, or [u]rl? " '(?b ?f ?u))))
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -1379,7 +1379,7 @@ Self-contained means that cross-referenced entries are merged."
              concat
              (format "  %s = {%s},\n" name value)))))
 
-(defun bibtex-completion-add-pdf-attachment ()
+(defun bibtex-completion-add-pdf-attachment (keys)
   ; this is basically a lowercase command alias for the function
   ; ideally, that name should be deprecated in favor of this one
   "Attach the PDFs of the entries with the given KEYS where available."
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -503,19 +503,23 @@ for string replacement."
   (bibtex-completion-init)
   (completing-read
    "BibTeX entries: "
-   (lambda (string pred action)
+   (lambda (string predicate action)
      (if (eq action 'metadata)
          '(metadata
+           ;; (annotation-function . bibtex-completion--annotation)
            (category . bibtex))
-       (complete-with-action action (bibtex-completion--get-candidates) string pred)))))
+       (complete-with-action action (bibtex-completion--get-candidates) string predicate)))))
 
 (defun bibtex-completion--get-candidates ()
   "Return all keys from bibtex-completion-candidates."
-  (mapcar
-   (lambda (cand)
-     (cons (bibtex-completion-format-entry cand (1- (frame-width)))
-           (cdr (assoc "=key=" cand))))
-   (bibtex-completion-candidates)))
+  (cl-loop
+   for candidate in (bibtex-completion-candidates)
+   collect
+   (cons
+    ;; Here use one string for display, and the other for search.
+    (propertize
+     (car candidate) 'display (bibtex-completion-format-entry candidate (1- (frame-width))))
+    (cdr (assoc "=key=" candidate)))))
 
 (defun bibtex-completion-candidates ()
   "Read the BibTeX files and return a list of conses, one for each entry.
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -72,7 +72,9 @@
     (define-key map (kbd "e") 'bibtex-completion-edit-notes)
     (define-key map (kbd "s") 'bibtex-completion-show-entry)
     (define-key map (kbd "l") 'bibtex-completion-add-pdf-to-library)
-    map))
+    map)
+  "Keymap for bibtex-completion commands"
+  )
 
 (defcustom bibtex-completion-bibliography nil
   "The BibTeX file or list of BibTeX files.
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -503,7 +503,11 @@ for string replacement."
   (bibtex-completion-init)
   (completing-read
    "BibTeX entries: "
-   (bibtex-completion--get-candidates)))
+   (lambda (string pred action)
+     (if (eq action 'metadata)
+         '(metadata
+           (category . bibtex))
+       (complete-with-action action (bibtex-completion--get-candidates) string pred)))))
 
 (defun bibtex-completion--get-candidates ()
   "Return all keys from bibtex-completion-candidates."
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -60,8 +60,8 @@
   :group 'completion)
 
 (defvar bibtex-completion-map
-  "Keymap for bibtex-completion commands"
   (let ((map (make-sparse-keymap)))
+    (define-key map (kbd "o") 'bibtex-completion-open-any)
     (define-key map (kbd "p") 'bibtex-completion-open-pdf)
     (define-key map (kbd "u") 'bibtex-completion-open-url-or-doi)
     (define-key map (kbd "c") 'bibtex-completion-insert-citation)
@@ -363,6 +363,10 @@ Only entries of these types are checked in order to resolve
 cross-references.  The default list is usually sufficient; adding
 more types can slow down resolution for large biblioraphies.")
 
+;; one can reset the backend function using fset; maybe there's a
+;; more elegant way to do this?
+(fset 'bibtex-completion-read-backend `bibtex-completion--completing-read)
+
 (defvar bibtex-completion-display-formats-internal nil
   "Stores `bibtex-completion-display-formats' together with the \"used width\" of each format string.
 This is set internally.")
@@ -379,7 +383,6 @@ bibliography file is reparsed.")
 (defvar bibtex-completion-string-hash-table nil
   "A hash table used for string replacements.")
 
-
 (defun bibtex-completion-normalize-bibliography (&optional type)
   "Return a list of bibliography file(s) in `bibtex-completion-bibliography'.
 If there are org mode bibliography-files, their corresponding
@@ -496,19 +499,22 @@ for string replacement."
 (defun bibtex-completion--read ()
   "Select BibTeX entries in completion system."
   ;; define a completion function that defaults to completing-read, but can be overridden
-  (list (cdr (assoc (bibtex-completion--completing-read) (bibtex-completion--get-candidates)))))
+  (list (bibtex-completion-read-backend)))
 
 (defun bibtex-completion--completing-read ()
   "Read bibtex-completion entries for completion using completing-read."
   (bibtex-completion-init)
-  (completing-read
-   "BibTeX entries: "
-   (lambda (string predicate action)
-     (if (eq action 'metadata)
-         '(metadata
-           ;; (annotation-function . bibtex-completion--annotation)
-           (category . bibtex))
-       (complete-with-action action (bibtex-completion--get-candidates) string predicate)))))
+  (when-let ((candidates (bibtex-completion--get-candidates))
+             (chosen
+              (completing-read
+               "BibTeX entries: "
+               (lambda (string predicate action)
+                 (if (eq action 'metadata)
+                     '(metadata
+                       ;; (annotation-function . bibtex-completion--annotation)
+                       (category . bibtex))
+                   (complete-with-action action candidates string predicate))))))
+    (cdr (assoc chosen candidates))))
 
 (defun bibtex-completion--get-candidates ()
   "Return all keys from bibtex-completion-candidates."
@@ -992,6 +998,7 @@ called in case no PDF is found."
 If multiple PDFs are found for an entry, ask for the one to open
 using `completion-read'.  If no PDF is found, try to open a URL
 or DOI in the browser instead."
+  (interactive (list (bibtex-completion--read)))
   (bibtex-completion-open-pdf keys 'bibtex-completion-open-url-or-doi))
 
 (defun bibtex-completion-format-citation-default (keys)


bibtex-completion.el
```diff
@@ -478,6 +493,23 @@ for string replacement."
 (defvar bibtex-completion-cached-notes-keys nil
   "A cache storing notes keys obtained when the bibliography was last parsed.")
 
+(defun bibtex-completion--read ()
+  "Select BibTeX entries in completion system."
+  ;; define a completion function that defaults to completing-read, but can be overridden
+  (list (cdr (assoc (completing-read-bibtex--read) (bibtex-completion--get-candidates)))))
+
+(defun completing-read-bibtex--read ()
```
bdarcus, line 501: Maybe this instead?

```suggestion
(defun bibtex-completion--completing-read ()
```

bibtex-completion.el
```diff
@@ -478,6 +493,23 @@ for string replacement."
...
+  "Select BibTeX entries in completion system."
+  ;; define a completion function that defaults to completing-read, but can be overridden
+  (list (cdr (assoc (completing-read-bibtex--read) (bibtex-completion--get-candidates)))))
+
+(defun bibtex-completion--completing-read ()
+  "Read bibtex-completion entries for completion using completing-read."
+  (bibtex-completion-init)
+  (completing-read
+   "BibTeX entries: "
+   (bibtex-completion--get-candidates)))
```
bdarcus, line 506: @minad - no rush, but how do I fix this to add back the completion category?

Here, for comparison was the original code from #355 that I started with, but I needed to simplify it a lot (I guess too much?) to get working here.

https://github.com/tmalsburg/helm-bibtex/pull/355/files#diff-43b58707c1adce5c6f3e1ed8d97ea5665182d94767472f7c196d630ff3ca0c3cR32-R37

bibtex-completion.el
```diff
@@ -478,6 +493,23 @@ for string replacement."
...
+
+(defun bibtex-completion--get-candidates ()
+  "Return all keys from bibtex-completion-candidates."
+  (mapcar (lambda (cand) (cons
+                          (bibtex-completion-format-entry cand (1- (frame-width))) (cdr (assoc "=key="                                                                                                    cand)))) (bibtex-completion-candidates)))
```
bdarcus, line 511: We need a better solution for this, as we're missing data in the process.

If we can't complete against the output of `bibtex-completion-candidates` using `completing-read`, can we transform it so we don't lose any of that data?

The use case here is I really want to add tags to complete against, so I can distinguish among different projects, etc. That works in `ivy-bibtex` and `helm-bibtex`, but not here.

See #362.

bdarcus:
Going to copy @myshevchuk here too, just in case he has time for input.

bibtex-completion.el
bdarcus: Attempted fix here.

https://github.com/tmalsburg/helm-bibtex/pull/361/commits/e5e4114691fec0dcca41d4577b7f16d44968049c

bibtex-completion.el
```diff
@@ -478,6 +493,30 @@ for string replacement."
...
+(defun bibtex-completion--completing-read ()
+  "Read bibtex-completion entries for completion using completing-read."
+  (bibtex-completion-init)
+  (completing-read
+   "BibTeX entries: "
+   (lambda (string pred action)
+     (if (eq action 'metadata)
+         '(metadata
+           (category . bibtex))
+       (complete-with-action action (bibtex-completion--get-candidates) string pred)))))
```
bdarcus, line 510: There appears to be something subtle wrong here, as I'm not getting the commands in `embark-act`.

bibtex-completion.el
```diff
@@ -478,6 +493,30 @@ for string replacement."
 (defvar bibtex-completion-cached-notes-keys nil
   "A cache storing notes keys obtained when the bibliography was last parsed.")
 
+(defun bibtex-completion--read ()
+  "Select BibTeX entries in completion system."
+  ;; define a completion function that defaults to completing-read, but can be overridden
+  (list (cdr (assoc (bibtex-completion--completing-read) (bibtex-completion--get-candidates)))))
```
bdarcus, line 499: Seems like the `bibtex-completion--get-candidates` call should be moved down into the `bibtex-completion--completing-read` function, given that it's only used there (it's not need in the ivy or helm packages).

bibtex-completion.el
minad: Sorry, @bdarcus I didn't follow closely what is going on here. I don't understand what you are attempting here, the PR by @mtreca has not been merged yet? The prototype by @mtreca using the completion category did work before, didn't it? The commit e5e4114 seems to be correct.

bdarcus: > I don't understand what you are attempting here, the PR by @mtreca has not been merged yet?

I'm providing this as an alternative PR, which @mtreca suggested, to see what modifying bibtex-completion itself would entail, and what impact it would have. 

I think it demonstrates what you were earlier arguing; that this is a better approach.

And it helps me understand better the code in doing so, regardless of which path we take.

> The prototype by @mtreca using the completion category did work before, didn't it?

Yes.

bibtex-completion.el
bdarcus: I think the answer is on the selectrum wiki:

https://github.com/raxod502/selectrum/wiki/Tips-for-Creating-Commands

bibtex-completion.el
```diff
@@ -478,6 +493,45 @@ for string replacement."
...
+
+(defun bibtex-completion--get-extra-search-data (candidate)
+  "Return extended search metadata as string."
+  ;; TODO
+  (if bibtex-completion-additional-search-fields
+      ; if the data is present, pull its value(s), join into a single string
+      ;(cl-loop
+      ; for field in (bibtex-completion-additional-search-fields)
+      ;(cdr (assoc field (cdr(candidate))))
+      ))
```
bdarcus, line 533: This is the last thing that needs to be resolved. It's just a simple function, but my elisp ...

bibtex-completion.el
```diff
@@ -478,6 +493,45 @@ for string replacement."
...
+    (cdr (assoc "=key=" candidate)))))
+
+(defun bibtex-completion--get-extra-search-data (candidate)
+  "Return extended search metadata as string."
+  ;; TODO
+  (if bibtex-completion-additional-search-fields
+      ; if the data is present, pull its value(s), join into a single string
+      ;(cl-loop
+      ; for field in bibtex-completion-additional-search-fields
+      ;(cdr (assoc field (cdr(candidate))))
```
bdarcus, line 532: This is the last thing that needs to be resolved. It's just a simple function, but my elisp ...

Not sure what should come after the `for field in ...` line.

bibtex-completion.el
bdarcus: I guess, @minad, the best way to debug this to come up with a minimal config file and use `emacs -Q`?

Any chance the selectrum et al dev community has come up with a standard minimal config for these purposes, perhaps that's posted on one of your wikis?

bdarcus: @oantolin - if you can, could you please take a look at this?

It should work, but embark isn't picking up the keybindings.

tmalsburg:
I'm dealing with a number of emergencies that require my full attention (nothing life threatening).  I will therefore mute this and related threads for a week or two.  This doesn't mean that I lost interest.  Thanks for your understanding.

bdarcus:
> I'm dealing with a number of emergencies that require my full attention (nothing life threatening). I will therefore mute this and related threads for a week or two. This doesn't mean that I lost interest. Thanks for your understanding.

Ok, I appreciate the head's up. Good luck with that!

In the meantime, I decided to merge @mtreca's code and mine here:

https://github.com/bdarcus/bibtex-actions

I strongly prefer to merge this PR and archive that repo, but it makes it a little easier to see the alternative, and use it now.

That code and this are functionally equivalent; just different command name prefixes.

bdarcus:
@mohkale - I just saw your note in the marginalia issue tracker that you were working on something similar. Have you seen this?

bibtex-completion.el
```diff
@@ -1321,6 +1379,13 @@ Self-contained means that cross-referenced entries are merged."
              concat
              (format "  %s = {%s},\n" name value)))))
 
+(defun bibtex-completion-add-pdf-attachment ()
+  ; this is basically a lowercase command alias for the function
+  ; ideally, that name should be deprecated in favor of this one
+  "Attach the PDFs of the entries with the given KEYS where available."
+  (interactive (list (bibtex-completion--read)))
+  (bibtex-completion-add-PDF-attachment))
```
mohkale, line 1387: There's an interactive argument here but the function doesn't accept any arguments :?

mohkale:
@bdarcus 

Oooh, nice to see someone adding this to bibtex propper.

I switched from ivy+counsel to selectrum+consult+orderless+embark a couple months back. Most of my bibtex stuff can be found [here](https://github.com/mohkale/emacs/blob/07e0c77d8b815bb4d0c3b31257613397b437fac4/init.org#consult-bibtex), it's mostly just an adaptation of `ivy-bibtex` but for the consult API.

I didn't open an issue for it on here yet cause everything is still pretty new and subject to change quite a bit. 

This issue seems to be trying to stay compatible with the `completing-read` API and not directly rely on consult/embark which I think is great. However I don't think I have much to contribute to it at this point.

bdarcus:
> This issue seems to be trying to stay compatible with the `completing-read` API and not directly rely on consult/embark which I think is great. 

Yes, that's right; at the strong suggestion of the author of consult, who participated in the linked discussions, and was a great help.

bibtex-completion.el
bdarcus: Oops; fixed.

Thanks.

mohkale:
@bdarcus 

> I'm wondering how to get nicer icon fonts to indicate PDFs, links, notes, and also how to filter on items that have those. See #363.

In regards to this, I'm not sure about how to approach this in a `completing-read` way but `consult` has a marvelous [narrowing](https://github.com/minad/consult#narrowing-and-grouping) feature where you can quickly discard a subset of completing-read entries interactively. For eg. with `consult-imenu` you can enter `f ` and it only shows function imenu entries. That may be useful for this.

bdarcus:
> `consult` has a marvelous [narrowing](https://github.com/minad/consult#narrowing-and-grouping) feature where you can quickly discard a subset of completing-read entries interactively. For eg. with `consult-imenu` you can enter `f ` and it only shows function imenu entries. That may be useful for this.

The cool thing about this approach is it introduces a ton of flexibility.

If this is merged, might be good to have a wiki page with config snippets and tips.

But that particular issue will get addressed with #363; it requires a little change to `bibtex-completion-candidates`.

bdarcus:
Converted this back to draft pending resolution of https://github.com/bdarcus/bibtex-actions/issues/17 (converting to `completing-read-multiple`).

bdarcus:
So here's what I'm thinking:

We should close PR for now.

Why?

I think the ideas embedded in it are sound, but for sure it would involve a change in the relations among the three packages here.

So I'm pretty sure you won't want to consider that for the short-term, and you may not even want to adopt it for the long-term.

Meanwhile, I've gotten comfortable enough with all of this to get much farther along on the independent project, which might benefit from that independence (with its own repo, issue tracker, discussion board, etc.).

OTOH, the code is here for later reference. So if you should want to incorporate the ideas in future releases, it will be easy enough to do. It's all pretty simple, after all, but there would be some decisions to make, on questions already raised (command names, for example, and whether worthwhile to refactor `helm-bibtex` and `ivy-bibtex`).

Let me know if I have that right when you get a chance @tmalsburg.

bdarcus:
A few other notes, from working more on the separate code:

1. I added some code to allow people to optionally use icon fonts for the pdf and note (and link, which I added) symbols, but I couldn't get bibtex-completion to use them, perhaps because these are propertized strings? I was wanting to experiment with whether it was possible to use the icon fonts, and ensure proper alignment with them. If the latter is true, it could be nice to allow this with bibtex-completion et al.
2. This is probably bigger change, and in the end I decided maybe not that important, but it might be nice if `bibtex-completion-format-entry` could take a config alist as input, so that more than one display string can be defined. This would be needed to support affixation config in Emacs 28, where you have separate strings for prefix, suffix, main candidate. I know ivy has a feature request in for this; not sure helm, and it could all be overkill in the end. Practically speaking, the only reason to do this is to have different formatting for different parts of the display.

On 1, the idea is one could do this:

``` elisp
(setq! bibtex-completion-pdf-symbol (all-the-icons-icon-for-file "foo.pdf" :face ;'all-the-icons-dred))
```

bdarcus:
I'm going to close this, and encourage instead merging the other smaller PRs I've submitted (though the title shortening one is the least important to me; just thought it might be helpful to someone).

As I said above, I do think the ideas here are worth considering longer term.

tmalsburg:
I see there is a separate package now (`bibtex-actions`) that implements a lot of the ideas discussed above.  Looks great and I'm looking forward to trying it out.

bdarcus:
Yep; as I dug into the code, I decided better to have a separate
project and instead submit small PRs to improve pieces.

linwaytin:
First thanks @tmalsburg  for this great package!!
I use it a lot when writing reports and articles.

Recently I switched from `ivy` to `selectrum` with `consult`.
The only function I haven't found a replacement is this package.
@mohkale has already had some nice code.
Do you mind to merge your code into this package or make it a separate package?
I really like to see this function so that I can totally get rid of the `ivy` dependence. (It's great. I just don't want to have too many packages on my system.)

bdarcus:
https://github.com/bdarcus/bibtex-actions

bdarcus:
> In regards to this, I'm not sure about how to approach this in a `completing-read` way but `consult` has a marvelous [narrowing](https://github.com/minad/consult#narrowing-and-grouping) feature where you can quickly discard a subset of completing-read entries interactively. For eg. with `consult-imenu` you can enter `f ` and it only shows function imenu entries. That may be useful for this.

@mohkale - I forgot about this post of your's above. How do you see this applying to bibtex-actions? On what would you narrow?

Daniel actually submitted a patch to emacs to add grouping support, which was merged yesterday.

mohkale:
@bdarcus 

> On what would you narrow?

You could narrow on what you specified (presence of PDF, or notes). I've actually never needed this but I just set it up so that you can narrow on the *kind* of bibtex entry it is.

```lisp
(require 'bibtex-completion)

(defvar consult-bibtex-history+ nil
  "History for `consult-bibtex+'.")

(defvar consult-bibtex-default-action+ 'consult-bibtex-insert-pdftools-link
  "The default action for the `consult-bibtex+` command.")

(defvar consult-bibtex-narrow+
  '((?b . "Book")
    (?u . "Unpublished")
    ;; Fallback for when none of the others apply.
    (?o . "Other"))
  "Narrowing configuration for `consult-bibtex+'.")

(defun consult-bibtex-candidates+ ()
  (cl-loop
   for cand in (bibtex-completion-candidates)
   with cand-str = nil
   do (setq cand-str
            (bibtex-completion-format-entry (cdr cand) (1- (frame-width))))
   ;; Add a `consult--type' property for narrowing support.
   do (add-text-properties 0 1
                           `(consult--type
                             ,(or
                               (when-let ((type (bibtex-completion-get-value "=type=" cand)))
                                 (car (rassoc (capitalize type) consult-bibtex-narrow+)))
                               (car (rassoc "Other" consult-bibtex-narrow+)))
                             consult--candidate
                             ,(bibtex-completion-get-value "=key=" cand))
                           cand-str)
   collect cand-str))

(defun consult-bibtex-read-entry+ (&optional arg local-bib)
  (when arg
    (bibtex-completion-clear-cache))
  (bibtex-completion-init)
  (let* ((candidates (consult-bibtex-candidates+))
         (preselect
          (when-let ((key (bibtex-completion-key-at-point)))
            (cl-find-if (apply-partially #'string-equal key) candidates))))
    (consult--read candidates
                   :prompt (format "BibTeX entries%s: " (if local-bib " (local)" ""))
                   :require-match t
                   :category 'bibtex-completion
                   :lookup 'consult--lookup-candidate
                   :default preselect
                   :title (consult--type-title consult-bibtex-narrow+)
                   :narrow (consult--type-narrow consult-bibtex-narrow+)
                   :history '(:input consult-bibtex-history+))))

(defun consult-bibtex+ (&optional arg local-bib)
  (interactive "P")
  (when-let ((entry (consult-bibtex-read-entry+ arg local-bib)))
    (if consult-bibtex-default-action+
        (funcall-interactively consult-bibtex-default-action+ entry)
      (warn "`consult-bibtex-default-action+' is unassigned."))))
```

Now if I run the command you can see entries of type book are grouped together and entries of type unpublished are also grouped together. I've set it up so anything without a type I'd like to narrow on gets put into the `Other` group.

Screenshot_20210521_114735

And I can use consult to narrow onto a specific group type such that none of the other candidates are visible.

Screenshot_20210521_115436

This is of course a very simplistic example. I can also set it up so that I can narrow on notes or pdfs (I plan to now). But as a demonstration it should be enough.

While narrowing is consult specific, you can group candidates with completing-read by using the metadata property.

```lisp
(completing-read "Prompt: "
                 (lambda (str pred action)
                   (if (eq action 'metadata)
                       `((x-title-function . GROUP-FUNCTION))
                     (complete-with-action action cands str pred))))
```

> Daniel actually submitted a patch to emacs to add grouping support, which was merged yesterday.

Incidentally I don't think the group-function minad submitted allows narrowing. All it does is let completion frontends like selectrum or bibtex group candidates together based on it, which could be enough to meet what you intended. Also I think this has been around for a while, minads recent changes were just changing the name of x-title-function to something like group-function (you can look that up I'm sure).

bdarcus:
> You could narrow on what you specified (presence of PDF, or notes).

That's what I was thinking.

So it's another way to handle what we now do with "initial-value" filters.

https://github.com/bdarcus/bibtex-actions/issues/43

>  I've actually never needed this but I just set it up so that you can narrow on the _kind_ of bibtex entry it is.

Got it.

> Incidentally I don't think the group-function minad submitted allows narrowing. All it does is let completion frontends like selectrum or bibtex group candidates together based on it, which could be enough to meet what you intended. Also I think this has been around for a while, minads recent changes were just changing the name of x-title-function to something like group-function (you can look that up I'm sure).

Yes; all correct. But it now formalizes grouping as a part of completing-read.

> I've added pdf and note based narrowing, now I can quickly see only the bibtex entries that have note files.

What do you think of the UX compared to the approach I took in bibtex-actions? Do you see an advantage?

mohkale:
@bdarcus 

> What do you think of the UX compared to the approach I took in bibtex-actions? Do you see an advantage?

I don't really use bibtex-completion, although I'm loving all the progress you've made with it so far. I feel like your implementation is better for regular commands. If users use `bibtex-actions-open-pdf` directly as an interactive command, instead of through embark, your filtering approach is clearly better. Users shouldn't be presented with bibtex entries that don't have PDFs when their trying to open PDFs. Arguably there could be some issues where a user thinks an entry does have a PDF when it doesn't, and they keep trying to find it through a filtered command only to realise it doesn't exist, but that's a configuration issue and I don't expect developers to account for all of those. The main advantage of narrowing (as described here) is that it's reversable, and the goal is flexible. You run one command to view all bibtex entries, then you quickly filter down to the ones you want. If you want to open PDFs you narrow to only files that don't have PDFs. If the one your after doesn't exist you can remove the narrowing and verify that it exists as an entry, but doesn't have a PDF. Personally I prefer the single entry point philosophy, one command, a million ways to use it :smile:. It's subjective.

bdarcus:
> The main advantage of narrowing (as described here) is that it's reversable, and the goal is flexible. You run one command to view all bibtex entries, then you quickly filter down to the ones you want. If you want to open PDFs you narrow to only files that don't have PDFs. If the one your after doesn't exist you can remove the narrowing and verify that it exists as an entry, but doesn't have a PDF. 

I'll have to experiment with that consult-based filtering.

We also tried to address this issue here: https://github.com/bdarcus/bibtex-actions/pull/120.

This uses a combination of "future" history and "predefined" search terms.

But I'm not sure if this was the best approach or not, how it compares to your approach, etc.

> Personally I prefer the single entry point philosophy, one command, a million ways to use it

I have thought about a single entry point, but not sure the point given the current design, given one could just write an alias to whatever you want and use that with embark.

oantolin:
> Personally I prefer the single entry point philosophy, one command, a million ways to use it 😄. It's subjective.

If you have a bunch of commands that work on the same type of objects, Embark let's you pick any single one of them and think of it as the "single entry point". For example, before writing Embark I used to use `find-file`, `delete-file`, `copy-file`, `rename-file`, etc. Now I think of `find-file` as the entry point and run the others as embark actions.

So for bibtex-actions, whichever command is the the one you use most often, think of that one as the entry point and then run the others as actions.

mohkale:
> So for bibtex-actions, whichever command is the the one you use most often, think of that one as the entry point and then run the others as actions.

@oantolin That's exactly what I do and meant by a millions "ways to use it". Apologies if that wasn't clear enough, although thanks for clarifying the point. :smile:.

oantolin:
Sorry, @mokhale, I think I read to quickly and didn't realize that's what you meant!

linwaytin:
> https://github.com/bdarcus/bibtex-actions

@bdarcus Can this package load the `.bib` file specified in a `.tex` file automatically, llike `ivy-bibtex-with-local-bibliography`? This is the function I'm using.

bdarcus:
> > https://github.com/bdarcus/bibtex-actions
> 
> @bdarcus Can this package load the `.bib` file specified in a `.tex` file automatically, llike `ivy-bibtex-with-local-bibliography`? This is the function I'm using.

I don't actually know.

I use bibtex-completion currently for file loading, but I'm not sure if that extends to this.

Do you know @mohkale?

bdarcus:
BTW, I've been active in helping get native org-cite functionality added to org-mode. This sort of feature would probably ideally be generalized to include org and also pandoc files, including csljson bib files.

mohkale:
> Do you know @mohkale?

Nope. I rarely use the in-file bibtex stuff and when I do it's only one or two of them. Is there a reason you don't want to create a propper bibtex file for them @linwaytin? In my case I keep everything I've ever read or intend to read in bibtex files because that makes it easier to reference them and pull up my notes for them. You can always make a local bibtex file for whatever your working on and use directory-local-variables to make `bibtex-completion` look in there as well.

oantolin:
I believe @linwaytin _is_ talking about using a bibtex file, and is wondering if bibtex-actions can figure out _which_ bibtex file it is by looking for a `\bibliography{use-this-one}` in the current buffer, instead of having to set a variable.

mohkale:
Oh, my bad, I read that too quickly :sweat:. 

I don't think it can, this seems more like like the kinda thing local-variables should be used for.

oantolin:
> I don't think it can, this seems more like like the kinda thing local-variables should be used for.

I disagree, looking at the current buffer to figure out which bibtex file to use seems like an obvious and great solution. Something  simple like this might be enough:

```elisp
(defvar old--bibtext-files nil)

(defun toggle-local-bibliography ()
  "Toggle whether to use the local bibliography."
  (interactive)
  (if old--bibtext-files
      (progn
        (setq bibtex-files old--bibtext-files
              old--bibtext-files nil)
        (message "Using global bibliography %s" bibtex-files))
    (save-excursion
      (goto-char (point-min))
      (if (search-forward-regexp "\\\\bibliography{\\([^}]+\\)}" nil t)
          (let ((bib (expand-file-name (concat (match-string 1) ".bib"))))
            (setq old--bibtext-files bibtex-files bibtex-files (list bib))
            (message "Using %s" bib))
        (message "No local bibliography found")))))
```

oantolin:
I think that advanced bibtex users tend to have a few large centralized bibtex files, with all the references they might use, maybe one file per subfield or something like that.; but many naive bibtex users (and I'm still at this stage!), tend to use one bib file per paper. The naive strategy seems inferior in the long run, but at least at first, the need to fix all the keys to be unique and keep things tidy seems overwhelming and it just feels easier to reduce the scope of the problem to the references of a single paper. People following the naive strategy tend to copy and paste references from the bib files of older papers into the new bib file, and then just search for the remaining references they need as they go along.

People using the one bib-file-per-paper approach probably don't even have a good setting for `bibtex-files` and would always want the local bibliography.

linwaytin:
Yes, I am not an advanced bibtex user. I currently use a bib file for a project. 
Thanks for all your answers.

bdarcus:
I played a bit with narrowing using `consult-buffer` @mohkale.

> The main advantage of narrowing (as described here) is that it's reversable, and the goal is flexible.

Yes, but so are initial-values and future history. One can simply delete the initial-value from the prompt, for example.

The narrowing UI is arguably a bit more elegant though.

mohkale:
> Yes, but so are initial-values and future history. One can simply delete the initial-value from the prompt, for example.

Yes. I seem to have misunderstood how that works there. I thought the collection was filtered before completing-read but it looks like its filtered afterwards in which case their mostly equivalent.

bdarcus:
> Yes. I seem to have misunderstood how that works there. I thought the collection was filtered before completing-read but it looks like its filtered afterwards in which case their mostly equivalent.

Here's how the "presets" feature works. If you add this ...

``` elisp
(setq bibtex-actions-presets '("book" "article" "inbook"))
```
... you can then access those by doing `M-n` from the prompt.

It doesn't, however, match against the type field, of course, but against the whole candidate string.

But the advantage is you could set those to a regexp to grab, for example, all articles published in the last 12 months, or whatever.

PR #366, opened 2021-03-20
Add =has-link= and bibtex-completion-link-symbol

### LINKED ISSUE #363: Add "=has-link=" to bibtex-completion search string
**EDIT: based on discussion below, I've narrowed the request, and updated title and this description accordingly.**

-------

I propose to add "=has-link" to the candidates, to indicate the presence of DOI or URL, and therefore those candidates where the "open" functions will return a result.

Could also add `has:note` etc to the candidate search strings instead of the symbols, which aren't ideal for search.

bdarcus:
Another small PR.

This is for indicating the presence of a DOI or URL link associated
with the entry, and therefore items for which the relevant "open" 
functions will return a result.

Closes #363.

Status: AFAIK, this should work, but doesn't appear to ATM. See [line comment](https://github.com/tmalsburg/helm-bibtex/pull/366#discussion_r603252555) below.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -183,6 +183,12 @@ This should be a single character."
   :group 'bibtex-completion
   :type 'string)
 
+(defcustom bibtex-completion-link-symbol "🔗"
+  "Symbol to indicate a DOI or URL link is available for a publication.
+This should be a single character."
+  :group 'bibtex-completion
+  :type 'string)
+
 (defcustom bibtex-completion-fallback-options
   '(("CrossRef                                  (biblio.el)"
      . (lambda (search-expression) (biblio-lookup #'biblio-crossref-backend search-expression)))
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -825,10 +825,14 @@ fields.  If FIELDS is empty, all fields are kept.  Also add a
 DO-NOT-FIND-PDF is non-nil, this function does not attempt to
 find a PDF file."
   (when entry ; entry may be nil, in which case just return nil
-    (let* ((fields (when fields (append fields (list "=type=" "=key=" "=has-pdf=" "=has-note="))))
+    (let* ((fields (when fields (append fields (list "=has-link" "=type=" "=key=" "=has-pdf=" "=has-note="))))
            ; Check for PDF:
            (entry (if (and (not do-not-find-pdf) (bibtex-completion-find-pdf entry))
                       (cons (cons "=has-pdf=" bibtex-completion-pdf-symbol) entry)
+                    entry))
+           ; Check for link:
+           (entry (if (or (assoc "doi" entry) (assoc "url" entry))
+                      (cons (cons "=has-link=" bibtex-completion-link-symbol) entry)
                     entry))
            (entry-key (cdr (assoc "=key=" entry)))
            ; Check for notes:
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -1324,7 +1324,7 @@ Self-contained means that cross-referenced entries are merged."
              unless (member name
                             (append (-map (lambda (it) (if (symbolp it) (symbol-name it) it))
                                           bibtex-completion-no-export-fields)
-                             '("=type=" "=key=" "=has-pdf=" "=has-note=" "crossref")))
+                             '("=type=" "=key=" "=has-pdf=" "=has-note=" "=has-link=" "crossref")))
              concat
              (format "  %s = {%s},\n" name value)))))
 
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -831,7 +831,7 @@ find a PDF file."
                       (cons (cons "=has-pdf=" bibtex-completion-pdf-symbol) entry)
                     entry))
            ; Check for link:
-           (entry (if (or (assoc "doi" entry) (assoc "url" entry))
+           (entry (if (or (cdr (assoc "doi" entry)) (cdr (assoc "url" entry)))
                       (cons (cons "=has-link=" bibtex-completion-link-symbol) entry)
                     entry))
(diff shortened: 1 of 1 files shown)


bibtex-completion.el
```diff
@@ -819,11 +825,15 @@ fields.  If FIELDS is empty, all fields are kept.  Also add a
...
 find a PDF file."
   (when entry ; entry may be nil, in which case just return nil
-    (let* ((fields (when fields (append fields (list "=type=" "=key=" "=has-pdf=" "=has-note="))))
+    (let* ((fields (when fields (append fields (list "=has-link" "=type=" "=key=" "=has-pdf=" "=has-note="))))
            ; Check for PDF:
            (entry (if (and (not do-not-find-pdf) (bibtex-completion-find-pdf entry))
                       (cons (cons "=has-pdf=" bibtex-completion-pdf-symbol) entry)
                     entry))
+           ; Check for link:
+           (entry (if (or (assoc "doi" entry) (assoc "url" entry))
```
bdarcus, line 834: I think this line is wrong.

tmalsburg:
Without having tested it, I think it doesn't work because the `=has-link=` tag is not included in `bibtex-completion-display-formats`.

Having said that, I'm not sure we necessarily need support for a `=has-link=` tag. Users who need to see entries with links can search for `http`. It's also not clear why there should be `=has-link=` but not `=has-isbn=`, `=has-keywords=`, and so on.  We could add tags for all possible fields but I'm afraid it would slow down things more that users with large bibliographies might like. (I'm perhaps also a bit biased against `=has-link=` because all my entries have links.)

bdarcus:
> I'm perhaps also a bit biased against `=has-link=` because all my entries have links.

I think this is the crux of it ;-)

Like half of my entries have doi or url links, but I'm a field (closer to the humanities) where I'm citing a lot of sources that are not journal articles and such.

The main user scenario is a user who wants to avoid running "open" against entries where nothing will be returned. So the justification is the tie to the commands/functions.

tmalsburg:
> Like half of my entries have doi or url links, but I'm a field (closer to the humanities) where I'm citing a lot of sources that are not journal articles and such.

Okay, I see.  I extrapolated from my field were everything I cite has a DOI.  I guess adding a link symbol makes sense then.  Let me know when you think the PR is ready to merge.

bdarcus:
> Let me know when you think the PR is ready to merge.

OK, thanks much. It's close.

To go back to why this doesn't appear to work, can you look at [this](https://github.com/tmalsburg/helm-bibtex/pull/366#discussion_r603252555) and tell me if it looks right?

It can't be because of the display-formats template, because this is before they're formatted.

Maybe it's related to the "additional-search-fields" variable? If yes, what should I be doing about that?

bibtex-completion.el
tmalsburg: Looks okay to me.  It should be true if either doi or url is defined.

tmalsburg:
> It can't be because of the display-formats template, because this is before they're formatted.

My suspicion was that `=has-url=` was actually added, just not displayed.

> Maybe it's related to the "additional-search-fields" variable? If yes, what should I be doing about that?

What is it that's not working?

bdarcus:
> What is it that's not working?

I'm looking at the content of `bibtex-completion-candidates`, which is where I intend this to be added. But this is what I see currently.

``` elisp
 ("The Bush Regime from Elections to Detentions https://papers.ssrn.com/abstract=942865 2007-05-01 Abraham, David report abraham2007"
  ("title" . "The Bush Regime from Elections to Detentions")
  ("url" . "https://papers.ssrn.com/abstract=942865")
  ("date" . "2007-05-01")
  ("author" . "Abraham, David")
  ("=type=" . "report")
  ("=key=" . "abraham2007"))
```

tmalsburg:
I see, hm.  Not sure what's going on, but check the function `bibtex-completion-prepare-entry` and especially the part after `Remove unwanted fields`.  Perhaps your extra field is added and then dropped at that point.

bibtex-completion.el
mohkale: @bdarcus

It looks to me like you're kind of not doing anything there

Expanding the expression out gives:

```lisp
           (entry (if (or (assoc "doi" entry)
                          (assoc "url" entry)))
```

Which is an if condition that has no body... so it always evaluates to nil. 

I beleive the intention was to check whether the entry has a doi or url field and if so add it to the `entry` alist. In which case this should work:

```lisp
           (entry (if-let ((link (cdr (or (assoc "doi" entry)
                                          (assoc "url" entry)))))
                      (cons (cons "=link=" link) entry)
                    entry))

```

mohkale: Oh wait, I believe you were trying to go for something like `=has-pdf=` where the value is a symbol to render. In which case this is probably what you want:

```lisp
           (entry (if-let ((link (cdr (or (assoc "doi" entry)
                                          (assoc "url" entry)))))
                      (cons (cons "=has-link=" bibtex-completion-link-symbol) entry)
                    entry))
```

bdarcus: So this is the current code:

``` elisp
           (entry (if (or (cdr (assoc "doi" entry)) (cdr (assoc "url" entry)))
                      (cons (cons "=has-link=" bibtex-completion-link-symbol) entry)
                    entry))
```

I intend that to be saying if the entry has a doi or url field, then add a "=has-link=" field (cons).

I was basing the code on the preceding one for pdfs, which obviously works, and was trying to be consistent with it.

mohkale: @bdarcus

Okay, looks good to me. What's the issue you wanted me to take a look at?

bdarcus: I was just hoping fresh eyes would see something we have not so far!

I'll take another close look soon and see if I can figure this out. Maybe at that point I'll have a more specific question.

Thanks for checking!

PR #367, opened 2021-03-24
Allow format-entry function to take optional template parameter

bdarcus:
Add:

- optional template parameter to `bibtex-completion-format-entry` 
- a new ` bibtex-completion-process-display-format` function

Remove:

- the `internal` format var and init setq (so to decouple  `bibtex-completion-format-entry` from  `bibtex-completion-init`)

Also, bump version to 1.1, so other packages can specify this requirement (since they would break if they relied on this change, but loaded an earlier version).

-----

Screenshot from 2021-03-26 11-04-25

I want to do as title says, without breaking anything of course ...

``` elisp
(setq myfoo-formats-suffix "(${=key=:15}) ${=type=:12}:${tags:30}")
```

The obvious and clean solution is to modify `bibtex-completion-format-entry` to take an optional template parameter.

The screenshot above is with this PR (also shows has-note filtering as mentioned in #363). The "suffix" content on the right is using a different face from the main display, and that content generated by a different template.

This involves a net neutral change in LOC.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -341,6 +341,14 @@ editor names."
   :group 'bibtex-completion
   :type '(alist :key-type symbol :value-type string))
 
+(defcustom bibtex-completion-display-formats-suffix nil
+  "Same format as display-formats template.
+When set, will render separate suffix string."
+  ; start with this; if needed, can always turn into a more general alt template
+  ; alist
+  :group 'bibtex-completion
+  :type '(alist :key-type symbol :value-type string))
+
 (defvar bibtex-completion-cross-referenced-entry-types
   '("proceedings" "mvproceedings" "book" "mvbook" "collection" "mvcollection")
   "The list of potentially cross-referenced entry types (in lowercase).
@@ -859,6 +867,7 @@ WIDTH is the width of the results list. The display format is
 governed by the variable `bibtex-completion-display-formats', or
 by ALT-DISPLAY-FORMATS if present."
   (let* ((format
+          ;; TODO: need to activate alt-display-formats somehow, bd
           (or (assoc-string (bibtex-completion-get-value "=type=" entry)
                             bibtex-completion-display-formats-internal
                             'case-fold)
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -360,6 +360,11 @@ more types can slow down resolution for large biblioraphies.")
   "Stores `bibtex-completion-display-formats' together with the \"used width\" of each format string.
 This is set internally.")
 
+
+(defvar bibtex-completion-display-formats-suffix-internal nil
+  "Stores `bibtex-completion-display-suffix-formats' together with the \"used width\" of each format string.
+This is set internally.")
+
 (defvar bibtex-completion-cache nil
   "A cache storing the hash of the bibliography content and the corresponding list of entries, for each bibliography file, obtained when the bibliography was last parsed.
 When the current bibliography hash is identical to the cached
@@ -419,12 +424,13 @@ Also sets `bibtex-completion-display-formats-internal'."
             (user-error "Bibliography file %s could not be found" file)))
             (bibtex-completion-normalize-bibliography))
 
-  ;; Pre-calculate minimal widths needed by the format strings for
-  ;; various entry types:
   (setq bibtex-completion-display-formats-internal
         (bibtex-completion-process-display-format
-         bibtex-completion-display-formats)))
-        ; BD: turn variable into alist?
+         bibtex-completion-display-formats))
+
+  (setq bibtex-completion-display-formats-suffix-internal
+        (bibtex-completion-process-display-format
+         bibtex-completion-display-formats-suffix)))
 
 (defun bibtex-completion-process-display-format (formats)
   "Pre-calculate minimal widths needed by the FORMATS strings for various entry types."
@@ -867,7 +873,8 @@ WIDTH is the width of the results list. The display format is
 governed by the variable `bibtex-completion-display-formats', or
 by ALT-DISPLAY-FORMATS if present."
   (let* ((format
-          ;; TODO: need to activate alt-display-formats somehow, bd
+          ;; TODO: need to activate
+          ;; bibtex-completion-display-formats-suffix-internal
           (or (assoc-string (bibtex-completion-get-value "=type=" entry)
                             bibtex-completion-display-formats-internal
                             'case-fold)
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -341,17 +341,6 @@ editor names."
   :group 'bibtex-completion
   :type '(alist :key-type symbol :value-type string))
 
-(defcustom bibtex-completion-display-formats-suffix
-    '((t . "${=has-pdf=:1}${=has-note=:1} ${=type=:7}"))
-  "For configuring a separate suffix display.
-This alist uses the same format as the display-formats template.  It can be
-used to render an affixation suffix or annotation, or simply a different
-face for this segment of the display."
-  ; start with this; if needed, can always turn into a more general alt template
-  ; alist
-  :group 'bibtex-completion
-  :type '(alist :key-type symbol :value-type string))
-
 (defvar bibtex-completion-cross-referenced-entry-types
   '("proceedings" "mvproceedings" "book" "mvbook" "collection" "mvcollection")
   "The list of potentially cross-referenced entry types (in lowercase).
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -364,7 +364,6 @@ bibliography file is reparsed.")
 (defvar bibtex-completion-string-hash-table nil
   "A hash table used for string replacements.")
 
-
 (defun bibtex-completion-normalize-bibliography (&optional type)
   "Return a list of bibliography file(s) in `bibtex-completion-bibliography'.
 If there are org mode bibliography-files, their corresponding
@@ -415,21 +414,28 @@ Also sets `bibtex-completion-display-formats-internal'."
   ;; Pre-calculate minimal widths needed by the format strings for
   ;; various entry types:
   (setq bibtex-completion-display-formats-internal
-        (mapcar (lambda (format)
-                  (let* ((format-string (cdr format))
-                         (fields-width 0)
-                         (string-width
-                          (string-width
-                           (s-format format-string
-                                     (lambda (field)
-                                       (setq fields-width
-                                             (+ fields-width
-                                                (string-to-number
-                                                 (or (cadr (split-string field ":"))
-                                                     ""))))
-                                       "")))))
-                    (-cons* (car format) format-string (+ fields-width string-width))))
-                bibtex-completion-display-formats)))
+        (bibtex-completion-process-display-format
+         bibtex-completion-display-formats)))
+        ; BD: turn variable into alist?
+
+(defun bibtex-completion-process-display-format (formats)
+  "Pre-calculate minimal widths needed by the FORMATS strings for various entry types."
+  (cl-loop
+   for format in formats
+   collect
+   (let* ((format-string (cdr format))
+          (fields-width 0)
+          (string-width
+           (string-width
+            (s-format format-string
+                      (lambda (field)
+                        (setq fields-width
+                              (+ fields-width
+                                 (string-to-number
+                                  (or (cadr (split-string field ":"))
+                                      ""))))
+                        "")))))
+     (-cons* (car format) format-string (+ fields-width string-width)))))
 
 (defun bibtex-completion-clear-cache (&optional files)
   "Clears FILES from cache.
@@ -846,12 +852,12 @@ find a PDF file."
   (cl-remove-duplicates entry
                         :test (lambda (x y) (string= (s-downcase x) (s-downcase y)))
                         :key 'car :from-end t))
-
 
-(defun bibtex-completion-format-entry (entry width)
+(defun bibtex-completion-format-entry (entry width &optional alt-display-formats)
   "Formats a BibTeX ENTRY for display in results list.
-WIDTH is the width of the results list.  The display format is
-governed by the variable `bibtex-completion-display-formats'."
+WIDTH is the width of the results list. The display format is
+governed by the variable `bibtex-completion-display-formats', or
+by ALT-DISPLAY-FORMATS if present."
   (let* ((format
           (or (assoc-string (bibtex-completion-get-value "=type=" entry)
                             bibtex-completion-display-formats-internal
@@ -884,7 +890,7 @@ governed by the variable `bibtex-completion-display-formats'."
               (- width (cddr format)))
             0 ?\s)))))))
 
-
+
 (defun bibtex-completion-clean-string (s)
   "Remove quoting and superfluous white space from BibTeX field value in S."
   (if s (replace-regexp-in-string "[\n\t ]+" " "
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -873,8 +873,6 @@ WIDTH is the width of the results list. The display format is
 governed by the variable `bibtex-completion-display-formats', or
 by ALT-DISPLAY-FORMATS if present."
   (let* ((format
-          ;; TODO: need to activate
-          ;; bibtex-completion-display-formats-suffix-internal
           (if alt-display-formats
               (or (assoc-string (bibtex-completion-get-value "=type=" entry)
                                 bibtex-completion-display-formats-suffix-internal
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -875,16 +875,16 @@ find a PDF file."
 WIDTH is the width of the results list. The display format is
 governed by the variable `bibtex-completion-display-formats', or
 by ALT-DISPLAY-FORMATS if present."
-  (let* ((format
-          (if alt-display-formats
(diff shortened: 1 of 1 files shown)


bibtex-completion.el
```diff
@@ -846,12 +852,12 @@ find a PDF file."
...
-(defun bibtex-completion-format-entry (entry width)
+(defun bibtex-completion-format-entry (entry width &optional alt-display-formats)
   "Formats a BibTeX ENTRY for display in results list.
-WIDTH is the width of the results list.  The display format is
-governed by the variable `bibtex-completion-display-formats'."
+WIDTH is the width of the results list. The display format is
+governed by the variable `bibtex-completion-display-formats', or
+by ALT-DISPLAY-FORMATS if present."
   (let* ((format
           (or (assoc-string (bibtex-completion-get-value "=type=" entry)
```
bdarcus, line 862: I assume this is where the key change needs to happen to enable the parameter.

bibtex-completion.el
```diff
@@ -846,17 +866,24 @@ find a PDF file."
...
+          ;; bibtex-completion-display-formats-suffix-internal
+          (if alt-display-formats
+              (or (assoc-string (bibtex-completion-get-value "=type=" entry)
+                                bibtex-completion-display-formats-suffix-internal
+                                'case-fold)
+                  (assoc t bibtex-completion-display-formats-suffix-internal))
+              (or (assoc-string (bibtex-completion-get-value "=type=" entry)
+                                bibtex-completion-display-formats-internal
+                                'case-fold)
+                  (assoc t bibtex-completion-display-formats-internal))))
```
bdarcus, line 886: I believe this core piece of code is correct, but I have to test it more. Specifically, it works for me with default (no suffix defined), but I need to test if the suffix generation actually works as expected.

bibtex-completion.el
```diff
@@ -352,6 +360,11 @@ more types can slow down resolution for large biblioraphies.")
   "Stores `bibtex-completion-display-formats' together with the \"used width\" of each format string.
 This is set internally.")
 
+
+(defvar bibtex-completion-display-formats-suffix-internal nil
```
bdarcus, line 364:
```suggestion
(defvar bibtex-completion--display-formats-suffix-internal nil
```
I was wanting to make these symbol name private using the `--` convention, but have not since it's not currently used anywhere else in the code. Would be nice to add this at some point.

bdarcus:
To update you @tmalsburg; here's what I've done on the PRs:

First, I closed #361, after concluding it didn't make sense now. It's easy enough to come back to at any point.

Instead, I submitted three small ones, all of them backward compatible; in order of importance:

1. this one
2. #366 to add `=has-link=`
3. #364 to add a shorten-title function; may or may not make sense

bibtex-completion.el
```diff
@@ -412,24 +427,32 @@ Also sets `bibtex-completion-display-formats-internal'."
             (user-error "Bibliography file %s could not be found" file)))
             (bibtex-completion-normalize-bibliography))
 
-  ;; Pre-calculate minimal widths needed by the format strings for
-  ;; various entry types:
   (setq bibtex-completion-display-formats-internal
```
bdarcus, line 430: I was also just wondering: couldn't this be set within the format-entry function?

bdarcus: I did a quick comparison of the existing bibtex-completion-format-entry, and my modified one.

``` elisp
ELISP> (benchmark-run 100
  (bibtex-actions--format-entry
   "aclu2011"
   (1- (frame-width))
   bibtex-actions-display-template))
(0.008320599000000001 0 0.0)

ELISP> (benchmark-run 100
  (bibtex-completion-format-entry
   "aclu2011"
   (1- (frame-width))))
(0.027423597 0 0.0)
```

So they're both fast, but using this approach is even faster.

Interestingly, my `bibtex-actions--get-candidates` function, which transforms the output of `bibtex-completion-candidates`, is _really_ slow by comparison, even though the UI in selectrum seems fast.

``` elisp
ELISP> (benchmark-run 100 (bibtex-completion-candidates))
(1.018316332 2 0.2629074330000023)

ELISP> (benchmark-run 100 (bibtex-actions--get-candidates))
(20.117633309 43 6.017887700999999)
```

bibtex-completion.el
```diff
@@ -24,7 +24,7 @@
 
 ;; A BibTeX backend for completion frameworks
 
-;; There are currently two fronends: helm-bibtex and ivy-bibtex.
+;; There are currently two frontends: helm-bibtex and ivy-bibtex.
```
bdarcus, line 27: now three though ;-)

bdarcus, line 27:
```suggestion
;; There are currently three frontends: helm-bibtex, ivy-bibtex and bibtex-actions.
```

bibtex-completion.el
```diff
@@ -4,7 +4,7 @@
 ;;         Justin Burkett <justin@burkett.cc>
 ;; Maintainer: Titus von der Malsburg <malsburg@posteo.de>
 ;; URL: https://github.com/tmalsburg/helm-bibtex
-;; Version: 1.0.0
+;; Version: 1.1.0
```
bdarcus, line 7: Actually, I'm a little unclear how MELPA versioning works.

If it's based on git tags, then since all these packages use the same repo, doesn't this need to be 2.1?

tmalsburg: Yes, that's my understanding as well (but I might be wrong).

tmalsburg:
Could you please briefly explain which problem this PR is supposed to address? And if the purpose is to format the entry list in various different ways, can't you just let-bind `bibtex-completion-display-formats` temporarily and then call the existing functions with that binding?  This way you can flexibly change the layout without changing anything in bibtex-completion.

bdarcus:
> ... if the purpose is to format the entry list in various different ways, 

That is indeed the purpose.

> can't you just let-bind `bibtex-completion-display-formats` temporarily and then call the existing functions with that binding? This way you can flexibly change the layout without changing anything in bibtex-completion.

I hadn't even considered that possibility. I'm kind of learning elisp as I go.

How would that work?

For comparison, here is what my [locally-modified](https://github.com/bdarcus/bibtex-actions/blob/78c782dd910abdc54a3dbc79b8d306ebf8c868b6/bibtex-actions.el#L124) `let*` code looks like, where "main" uses one template and "suffix" uses another:

``` elisp
   (let* ((pdf (if (assoc "=has-pdf=" (cdr candidate)) " has:pdf"))
          (note (if (assoc "=has-note=" (cdr candidate)) "has:note"))
          (link (if (assoc "doi" (cdr candidate)) "has:link"))
          (citekey (bibtex-completion-get-value "=key=" candidate))
          (candidate-main
           (bibtex-actions--format-entry
            candidate
            (1- (frame-width))
            bibtex-actions-template))
          (candidate-suffix
           (bibtex-actions--format-entry
            candidate
            (1- (frame-width))
            bibtex-actions-template-suffix))
```

tmalsburg:
Temporarily let-binding configuration variables is an elisp technique that I also needed to get used to.  Nothing that I was familiar with from other languages.  It allows for a lot of flexibility but can also become messy and result in code that's difficult to read and debug.

Specifically, what I was think is something like this:

``` elisp
(let ((bibtex-completion-display-formats (your desired format)))
  (bibtex-completion-format-entry entry width))
```

With this setup, the entry will be formatted not with the globally configured layout but with the one specified in the let-binding.

bdarcus:
> Specifically, what I was thinking of is something like this:
> 
> ```emacs-lisp
> (let ((bibtex-completion-display-formats (your desired format)))
>   (bibtex-completion-format-entry entry width))
> ```

I see how there that would redefine the template.

But can I run the format-entry function against two different templates in the same process, each of them user customizable?

That's what I'm doing now, where I adapted the format-entry function to require the template in my code, with the intention to remove it if you merged this.

Or is what you're thinking here to have a kind of format-entry wrapper that does that let-binding? Like:

``` elisp
(defun bibtex-actions--format-entry (entry width template)
  ;; here do the let-binding
  (bibtex-completion-format-entry ...)
```

bdarcus:
> So is what you're thinking here to have a kind of format-entry wrapper that does that let-binding?

Hmm ... would that even work though, given that now there's `bibtex-completion-display-formats-internal` in the init function?

tmalsburg:
> So is what you're thinking here to have a kind of format-entry wrapper that does that let-binding? Like:

To be honest, I'm not sure what exactly you're trying to achieve.  I was thinking that you want to render the two halves of the table (left and right) separately.  My idea was that you could call `bibtex-completion-format-entry` two times, once for the left-hand side and once for the right-hand side, and then concatenate the two halves to form the complete table. But perhaps I'm misunderstanding what you'd like to achieve.

bdarcus:
No, you're understanding is correct.

I'll play with it when I get a chance.

tmalsburg:
Can't you render the table in one go and then apply font styling for each half each line separately?

bdarcus:
> Can't you render the table in one go and then apply font styling for each half each line separately?

That's another idea I didn't consider.

But wouldn't that be more complicated?

Concatenatng two template strings is simple, of course.

But how do I know where to attach the face (I just have one for the "suffix" section) to the string?

And what happens if one or both templates are alists rather than strings?

tmalsburg:
> But how do I know where to attach the face (I just have one for the "suffix" section) to the string?

I'm not super familiar with faces but I would guess that you can assign faces to a ranges of strings. So you'd have to assign one fact to the first n characters and the another to the remainder of the line. Should be too difficult.

> And what happens if one or both templates are alists rather than strings?

I'm not sure what you mean.  It would help to see some sketch in code.

bdarcus:
> I would guess that you can assign faces to a ranges of strings. So you'd have to assign one fact to the first n characters and the another to the remainder of the line.

IC. So grab n from the "display-formats-internal" variable.

Does it matter more generally (beyond faces) that that n would only represent one template?

> I'm not sure what you mean. It would help to see some sketch in code.

It's not a huge deal; I just mean, say the user has this in their config:

``` elisp
    '((article       . "${year:4} ${author:36} ${title:*} ${journal:40}")
      (inbook        . "${year:4} ${author:36} ${title:*} Chapter ${chapter:32}")
      (incollection  . "${year:4} ${author:36} ${title:*} ${booktitle:40}")
      (inproceedings . "${year:4} ${author:36} ${title:*} ${booktitle:40}")
      (t             . "${year:4} ${author:36} ${title:*}")))

(setq bibtex-actions-suffix-format '((t . "${=has-pdf=:1}${=has-note=:1} ${=type=:7}")))
```

That would require some logic to pull the right "display-formats" template for the entry to concatenate with the "suffix" template. 

On reflection, it's not complicated, but it is additional code to write.

I do have this all working well with my forked format-entry function (which requires the template argument), and the logic is straightforward.

I guess we've established that I have options, though, including just keeping my code as is.

What's your concern about merging this? That you don't see the need, and worry about something breaking?

tmalsburg:
I don't see why `bibtex-actions-suffix-format` is needed.  Can't you simply integrate the suffix into `bibtex-completion-display-formats`?  The idea being that bibtex-completion renders the lines for you and then your own code adds the faces which are specific to your package and not needed in helm/ivy-bibtex. 

> What's your concern about merging this? That you don't see the need, and worry about something breaking?

Multiple things:  I don't see the need, within bibtex-completion at least. It seems more natural to have the code for styling the lines with faces withing your package where it's actually needed.  The added complexity will also make it harder to maintain this code base (which is already challenging).

bdarcus:
> Multiple things: I don't see the need, within bibtex-completion at least. It seems more natural to have the code for styling the lines with faces withing your package where it's actually needed. The added complexity will also make it harder to maintain this code base (which is already challenging).

I'm totally fine with you closing this. 

My code is licensed same as your's, so if I do something useful you want to borrow at some point, that will be easy to do.

But to offer counterpoints to consider:

1. Sure, but the formatting function is hard-coded to accept only a single template, which gives third-party front-ends like mine a lot less flexibility, without any obvious advantage.
2. I'm biased of course, but I think the changes make it easier to understand and maintain; it's actually simpler, I would say. Here, formatting is entirely contained within the formatting function, without any dependence on some global variable that's created by the init function. The init function also become simpler.

But as I say, I won't take it personally if you disagree, and appreciate you having found the time to look into these ideas I've submitted!

tmalsburg:
> Sure, but the formatting function is hard-coded to accept only a single template, which gives third-party front-ends like mine a lot less flexibility,

I must be missing something because, as far as I understand, it should be easy to achieve the layout you show in the screenshot with the current facilities (by combining the main and suffix template and then apply faces in a second step).  I don't see what is gained by adding the ability to supply multiple template.

> it's actually simpler, I would say. Here, formatting is entirely contained within the formatting function, without any dependence on some global variable that's created by the init function. The init function also become simpler.

I agree that the precalculation isn't super elegant but there was a reason to do it that way: If we precaculate the formatting at init, we do it just once.  If we calculate the formatting in `bibtex-completion-format-entry` (as I think this PR does), it will be done for every single entry which would be slower.

bdarcus:
> If we precaculate the formatting at init, we do it just once. If we calculate the formatting in `bibtex-completion-format-entry` (as I think this PR does), it will be done for every single entry which would be slower.

Yes, that's true.

But OTOH, my benchmarking shows this runs a bit faster (though I assume that might be because of cl-loop?).

bdarcus:
On this:

> it should be easy to achieve the layout you show in the screenshot with the current facilities (by combining the main and suffix template and then apply faces in a second step). I don't see what is gained by adding the ability to supply multiple template.

That may work in this particular circumstance, but:

1. I don't see why I should be forced into that particular choice. It's easier for me to continue using my forked format-entry function instead at this point.
2. where I started with this, this approach wouldn't have worked.

So, for example, if someone wants to implement the suffix-like content using affixation and/or annotation (which is what I originally did), you do need multiple templates. That's where this PR really came from, and a number of people I've seen start to do something like what I have done have wanted to use affixation and/or annotation. Either way, you need multiple templates to do that.

But on reflection, even for my case, I still don't think the format-entry function should hard-code a template.

Hence why I still think this is a good idea :-).

bdarcus:
> ~I could always move that back though.~ (actually, I don't think I can)

I guess it would be possible, but then the template parameter would need to be required, and take the processed template?

bdarcus:
A potentially bigger problem, though I am unsure how big:

https://github.com/bdarcus/bibtex-actions/issues/67

Obviously wouldn't happen if I followed your suggested approach, @tmalsburg, but can you help me understand what would be required to fix that?

Basic issue is the math for format-entry assumes one template too, so that field width don't work entirely correctly if more than one.

tmalsburg:
> ... whereas before the change, it was closer to a minute for the run of 100.

Formatting 100 entries should under no circumstances take a minute.  On my system and with the current code I can format thousands of entries in a fraction of a second.  Am I missing something?

tmalsburg:
> Obviously wouldn't happen if I followed your suggested approach, @tmalsburg, but can you help me understand what would be required to fix that?

The lecture period is starting next week and I'm moving to a new university and city (with my family).  No time currently, sorry. 

> BTW, what is the purpose of the beginning "t" in this?

You can have separate templates for different types of entries.  The template with `t` is the default template.  Here's my own config as an example:

``` elisp
    '((article       . "${author:30} ${title:*} ${journal:20} ${year:4} ${=has-pdf=:1}${=has-note=:1} ${=type=:3}")
      (inbook        . "${author:30} ${chapter:*} ${title:20} ${year:4} ${=has-pdf=:1}${=has-note=:1} ${=type=:3}")
      (incollection  . "${author:30} ${title:*} ${booktitle:20} ${year:4} ${=has-pdf=:1}${=has-note=:1} ${=type=:3}")
      (inproceedings . "${author:30} ${title:*} ${booktitle:20} ${year:4} ${=has-pdf=:1}${=has-note=:1} ${=type=:3}")
      (phdthesis     . "${author:30} ${title:*} ${school:20} ${year:4} ${=has-pdf=:1}${=has-note=:1} ${=type=:3}")
      (t             . "${author:30} ${title:*} ${year:4} ${=has-pdf=:1}${=has-note=:1} ${=type=:3}")))

```

tmalsburg:
> I'm trying to figure out how to implement your suggestion on the linked PR, but stuck at merging the templates, particularly as I contemplate ones where there are multiple entry types. Seems like fixing the above may be easier.

Very quickly, another idea I had:  You could format the left half of the table and then

bdarcus:
Not sure.

Maybe the benchmark-run isn't an accurate reflection of real world
performance.

In practice, with the cache added and moving template calculation (and one
or two other things) to the calling function, bibtex-actions performance is
now perceptually the same as ivy-bibtex and helm-bibtex; a slight pause on
initial load with larger libraries, and then instant access when running
off the cache.

But my point was your observation above about the recomputation of the
template each format-entry run was definitely on the right track.

tmalsburg:
> But my point was your observation above about the recomputation of the template each format-entry run was definitely on the right track.

Ah, I see.  I misinterpreted "the change".  Sorry. As I said, I'm a bit overwhelmed currently.

bdarcus:
In any case, if you were to merge this at some point, it would just require:

1. change a single line in `helm-bibtex` and `ivy-bibtex` to include the processed template as an argument
2. to remove the processing of that template on the `format-entry` function
3. to bump the versions of all three packages, and update the `bibtex-completion` dependency on the front-ends

`Bibtex-actions` already implements this, so you can always just borrow from that.

tmalsburg:
@bdarcus thank you!

PR #376, opened 2021-06-08
Add pixel alignment of bibtex fields

ilupin:
As shown below, when bibtex fields contain multi-byte characters, such as CJK characters, the fields are not aligned.

image

To fix the problem, this PR adds the text property `:align-to` to the end of field. 

image

This PR also works when variable-pitch face is used.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -850,6 +850,20 @@ find a PDF file."
                         :key 'car :from-end t))
 
 
+(defun bibtex-completion-format-field (field-name entry)
+  "Return Formatted value of FIELD-NAME for ENTRY."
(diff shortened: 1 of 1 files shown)


tmalsburg:
Wow, nice work! I'll have a closer look at the implementation in the next days and will then likely merge.

Side note: Table formatting is such a common problem in Emacs, I wish there was an API for that.

ilupin:
Github always showed it in Recent activity on the main page, so I closed it to hide the reminder. The code has no problem.

tmalsburg:
It's really cool that you found a solution for aligning the columns even when multi-byte characters are used. Would you mind explaining your approach in a sentence or two.  I would like to make sure I understand what's going on before I merge.  Thank you.

Issue #381, opened 2021-07-15
[Feature request/Question] A possibly reverse bibtex-completion-find-pdf?


c1-g:
Is there any function that takes a filepath, then return an entry key for it as a string? We could generalize this further by say, `bibtex-completion-find-key-with-field (string &optional field)`, this function takes `STRING` then match it against all fields of every bibtex entry in `bibtex-completion-bibliography` then return a list of keys with `STRING` in their fields, or to be even more specific, match `STRING` against `FIELD` of  every bibtex entry, return a list of keys that matches. I can work on a pull request as this feature will make my note-taking even more convenient and open a lot of possibilities for writing some extensions in the future but, I have no clue how to do this since the relavant bibtex-completion functions seem to take a key as their argument.

tmalsburg:
> Is there any function that takes a filepath, then return an entry key for it as a string?

Well, you can add the `file` field to `bibtex-completion-additional-search-fields` and then search for file paths using the usual UI.

> We could generalize this further by say, bibtex-completion-find-key-with-field (string &optional field), this function takes STRING then match it against all fields of every bibtex entry in bibtex-completion-bibliography then return a list of keys with STRING in their fields, or to be even more specific, match STRING against FIELD of every bibtex entry, return a list of keys that matches.

That’s the usecase for which I designed helm/ivy-bibtex :)  What’s wrong with the current approach?

> I can work on a pull request as this feature will make my note-taking even more convenient and open a lot of possibilities for writing some extensions in the future but

Interesting.  Could you please describe your use case in more detail?  How exactly would the function that you’re proposing make note-taking more efficient?  Perhaps there already are facilities to solve your specific problem.

> I have no clue how to do this since the relavant bibtex-completion functions seem to take a key as their argument.

You can access a list of all entries via `bibtex-completion-candidated`.  It’s then relatively easy to filter them.

tmalsburg:
I will have a closer look at your PR when I have a better understanding of what you're trying to achieve.

PR #382, opened 2021-07-16
bibtex-find-key-from-file function

c1-g:
The initial idea is from #381. Their docstrings are quite extensive.
I've written four functions and one variable,
the function `find-key-from-file` delegates finding key to every function in the new variable `find-key-functions` each takes one argument called FILENAME.

The `find-key-functions` includes
1. `find-key-from-note`, a function that returns a list of key associated with FILENAME and `bibtex-completion-notes-path` depending on its value.

2. `find-key-from-pdf`, thats use another new function `get-entry-with-string` (See description in #381. TLDR; list every entry with STRING in its fields.) to return a list of entries with FILENAME in its fields then extract the `"=key="` field from each entry in the list or just return the BibTex part of FILENAME in a list.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -694,6 +694,118 @@ Fields from crossreferenced entries are appended to the requested entry."
           (display-warning :warning (concat "Bibtex-completion couldn't find entry with key \"" entry-key "\"."))
           nil)))))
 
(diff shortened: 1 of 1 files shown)


c1-g:
https://github.com/tmalsburg/helm-bibtex/issues/381 Answers has solve my problem! Thanks!

Issue #383, opened 2021-07-24
Support non-PDF files (JPG, PDF, HTML, ePub, directory)


salutis:
_Note: I was told to ask here, see https://github.com/bdarcus/bibtex-actions/issues/188._

I have, for instance, the following `.bib` entry:

```
@Misc{monet-1877a,
  author       = {Claude Monet},
  title        = {La Gare Saint-{Lazare}},
  year         = 1877,
  type         = {painting},
  publisher    = {Musée d'Orsay, Paris, France}
}
```

The `.bib` entry has a corresponding `monet-1877a.jpg` file.

When I use `bibtex-actions-open`, instead of seeing the painting, I am informed that:

```
Wrong type argument: stringp, nil
```

I would like to open JPG, PDF, HTML, ePub, or even directories, that match the citation keys.

bdarcus:
I was suggesting, @tmalsburg, that ideally you's support other file types, or if not, provide error handling for wrong file types.

salutis:
> I was suggesting, @tmalsburg, that ideally you'd support other file types, or if not, provide error handling for wrong file types.

Ideally, this would be _any kind_ of file (including directories) that matches the key. For instance, if one cites `foo-2020` and there exists `foo-2020.mp3`, why not open it, or at least reveal it in `dired`? People cite images, videos, and more; not just PDF documents.

tmalsburg:
Your use case makes sense, but here's a problem when we allow any file type: Many users store their note files along with the PDFs and then have `Authors2020.pdf` and `Authors2020.org` in the same directory.  If those users select "Open PDF, URL, or DOI", they will be asked every time whether they'd to open the note file or the PDF which is unnecessary.

Two possible solutions:

1. You could list the .jpg in the file field.  My memory is that we allow any file type there:

``` bibtex
@Misc{monet-1877a,
  author       = {Claude Monet},
  title        = {La Gare Saint-{Lazare}},
  year         = 1877,
  type         = {painting},
  publisher    = {Musée d'Orsay, Paris, France},
  file = {monet-1877a.jpg}
}
```

2. You could define a custom function for opening JPGs.  `bibtex-completion-open-url-or-doi` could be used as a template for that function.

salutis:
[...]

PR #384, opened 2021-07-26
README: Mention JPEG in non-PDF formats alongside DjVu

salutis:
Stems from https://github.com/tmalsburg/helm-bibtex/issues/383

This patch is to clarify the fact that non-PDF-like files are perfectly OK, despite the variable being called `bibtex-completion-pdf-extension`.


tmalsburg:
Thanks.

Issue #306, opened 2020-02-03
ivy-bibtex Apply actions to multiple entries lead to error: command attempted to use minibuffer while in minibuffer


atanasj:
When trying to enter multiple citations I get the error `command attempted to use minibuffer while in minibuffer` and cannot enter any citations. However, when entering a single citation, it works as expected. 

My config is: 

```emacs-lisp
(use-package ivy-bibtex
  :config
  (setq bibtex-completion-bibliography '"~/.pandoc/MyLib.bib")
  (setq bibtex-completion-display-formats
        '((article       . "${=has-pdf=:1}${=has-note=:1} ${=type=:3} ${year:4} ${author:36} ${title:*} ${journal:40}")
          (inbook        . "${=has-pdf=:1}${=has-note=:1} ${=type=:3} ${year:4} ${author:36} ${title:*} Chapter ${chapter:32}")
          (incollection  . "${=has-pdf=:1}${=has-note=:1} ${=type=:3} ${year:4} ${author:36} ${title:*} ${booktitle:40}")
          (inproceedings . "${=has-pdf=:1}${=has-note=:1} ${=type=:3} ${year:4} ${author:36} ${title:*} ${booktitle:40}")
          (t             . "${=has-pdf=:1}${=has-note=:1} ${=type=:3} ${year:4} ${author:36} ${title:*}")))
  (setq bibtex-completion-additional-search-fields '(keywords))
  (setq bibtex-completion-pdf-symbol "⌘")
  (setq bibtex-completion-notes-symbol "✎")
  (setq bibtex-completion-format-citation-functions
        '((org-mode      . bibtex-completion-format-citation-org-link-to-PDF)
          (latex-mode    . bibtex-completion-format-citation-cite)
          (markdown-mode . bibtex-completion-format-citation-pandoc-citeproc)
          (default       . bibtex-completion-format-citation-default)))
  (setq ivy-bibtex-default-action 'ivy-bibtex-insert-citation)
  )
```

Below is the associated debugger output:

```emacs-lisp
Debugger entered--Lisp error: (error "Command attempted to use minibuffer while in minibuffer")
[...]

PR #388, opened 2021-08-17
Ivy mark cands

franburstall:
This adds marking and acting on candidates to `ivy-bibtex` as requested in #306.

`Ivy` has supported this for some time now so only minor changes are needed:
1. Adjust `ivy-bibtex-display-transformer` so that marks are visible: d10a608e62e6b4862ce601e301bd58947f228510
2. Provide multi-actions that act on lists of marked candidates.  Without these, `ivy` simply applies the chosen action to each candidate in turn but `bibtex-completion-*` already know how to do DTRT with lists of keys: 081ade9d509d489207ce05ff52e121e99d24eb83  https://github.com/tmalsburg/helm-bibtex/commit/a584b8399aa8047943d58a6764beb115e1789b0c https://github.com/tmalsburg/helm-bibtex/commit/83e9a691b69245ccbd1db4319dea45f2da1127fa
3. `Ivy` only binds `ivy-toggle-marks` by default unless `ivy-hydra` is loaded.  We provide convenient bindings to mark individual candidates: https://github.com/tmalsburg/helm-bibtex/commit/102a7d6a6c3b508f5cfb24ce88476553989187d3

I am happy to update the documentation if you are minded to accept this PR.

Thank you!

### DIFF
--- a/README.org
+++ b/README.org
@@ -7,7 +7,7 @@ Ivy-bibtex: [[http://melpa.org/#/ivy-bibtex][http://melpa.org/packages/ivy-bibte
 Helm-bibtex and ivy-bibtex allow you to search and manage your BibTeX bibliography.  They both share the same generic backend, bibtex-completion, but one uses the Helm completion framework and the other Ivy as a front-end.
 
 * News
-- 2021-08-25: It is now possible to mark and act on multiple entries in ~ivy-bitex~.  See [[#multi-actions][here]].
+- 2021-08-25: It is now possible to mark and act on multiple entries in ~ivy-bitex~.  See [[#apply-actions-to-multiple-entries][here]].
 - 2021-07-25: ~helm-bibtex-with-local-bibliography~ and ~ivy-bibtex-with-local-bibliography~ now also use locally and globally defined bibliographies in org files.  These are bibliographies specified using the new ~#+BIBLIOGRAPHY:~ key word and those in the variable ~org-cite-global-bibliography~.
 - 2021-07-18: Added a citation function for Org’s new citation system: ~bibtex-completion-format-citation-org-cite~ (for use in configuration variable ~bibtex-completion-format-citation-functions~)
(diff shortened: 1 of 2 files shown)


tmalsburg:
Thanks for the PR. Useful feature but two coments:

1. Isn't it possible to have just one version of each action function that works with a single candidate and with multiple candidates? I think this could simplify the code.
2. Can we get rid of the additional defcustoms? This package already has a ridiculous amount of defcustoms (my fault) and I'd prefer reasonable defaults over configurability.

franburstall:
> 1. Isn't it possible to have just one version of each action function that works with a single candidate and with multiple candidates? I think this could simplify the code.

Yes, I will do that.  It will not simplify too much as we still have to provide two functions (or the same function twice) to 'ivy'.

> 2. Can we get rid of the additional defcustoms? This package already has a ridiculous amount of defcustoms (my fault) and I'd prefer reasonable defaults over configurability.

Sure.  Is it OK to replace them with defvars?  I can imagine people wanting to adjust `ivy-bibtex-default-multi-action` and, perhaps, `ivy-bibtex-use-extra-keymap`.

tmalsburg:
> Yes, I will do that. It will not simplify too much as we still have to provide two functions (or the same function twice) to 'ivy'.

Ivy makes this more complicated than necessary, doesn't it?

> Is it OK to replace them with defvars?

Sure.

Finally, it would be good to have some documentation in README.org and a news item at the top of the document.

franburstall:
> Finally, it would be good to have some documentation in README.org and a news item at the top of the document.

OK.  This might take a day or two.

franburstall:
> > Finally, it would be good to have some documentation in README.org and a news item at the top of the document.
> 
> OK. This might take a day or two.

In the process of doing this, I find several links to a non-existing section *Action for opening annotated PDFs* with custom_id *annotated*.  Shall I fix these while I am in there?

tmalsburg:
Merged. Thank you.  Not sure what's up with that broken link.

PR #390, opened 2021-09-24
Update ivy-bibtex.el

jkitchin:
There is a bug in `ivy-bibtex-display-transformer` that causes marking candidates to stop working. The issue is the space that gets inserted. It should be an empty space.


tmalsburg:
Merged.  Thank you, John.

PR #401, opened 2022-01-08
Add function bibtex-completion-get-key-org-cite

akirakyle:
Adding support for the new org cite syntax. This then allows one to trivially use `helm-bibtex` or `ivy-bibtex` as an `org-cite-follow-processor' by setting

```
(org-cite-register-processor 'ivy-bibtex-org-cite-follow
  :follow (lambda (_ _) (ivy-bibtex)))

(setq org-cite-follow-processor 'ivy-bibtex-org-cite-follow)
```

While this doesn't handle follow when point is on `cite:` as there may be multiple keys, it's good enough for me.

### DIFF
--- a/README.org
+++ b/README.org
@@ -727,6 +727,19 @@ Helm-bibtex and ivy-bibtex display entries in the order in which they appear in 
             :filter-return 'reverse)
 #+END_SRC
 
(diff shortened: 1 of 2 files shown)


akirakyle:
I added a section to README.org under Advanced usage. I wasn't sure if this is the right place or if I used the right language so feel free to change it as you see fit. Thanks!

akirakyle:
Thanks!

PR #407, opened 2022-07-06
add a separate follow processor for helm

mpedramfar:
Thanks for your great package!

Having helm-bibtex open as the follow processor is not ideal, since to take an action, we need to press tab and then not all actions are meaningful when point is on a citation. The PR adds a separate follow processor and two new variables, `helm-bibtex-follow-actions-alist` and `helm-bibtex-follow-full-frame`, to customize it. To use this processor, this should be added to the init file:
```elisp
(org-cite-register-processor 'my-bibtex-org-cite-follow
  :follow 'helm-bibtex-follow)

(setq org-cite-follow-processor 'my-bibtex-org-cite-follow)
```

### DIFF
--- a/README.org
+++ b/README.org
@@ -137,6 +137,16 @@ The names of these files consist of the BibTeX key plus a user-defined suffix (~
 
 At this point most people will be ready to go.  Skip to [[#usage][Usage]] below to see how to use helm-bibtex and ivy-bibtex.
 
+** Follow processor for helm
+
+Invoking ~helm-bibtex~ or ~ivy-bibtex~ when point is on an [[https://orgmode.org/manual/Citation-handling.html][org-mode citation]] will automatically select that key. However, the default ~org-open-at-point~ on a org citation will take you to the corresponding bibliography entry. The following code will change this behavior to instead open ~helm-bibtex-follow~ when following an org citation by entering ~RET~ or clicking on it:
+
+#+BEGIN_SRC elisp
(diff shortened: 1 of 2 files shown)


tmalsburg:
Hi and thank you.  I can't test this right now.  Could you please explain in one sentence what this follow processor does differently?  And woudn't it be better to just replace the existing processor?

mpedramfar:
Sure, but it's more than 1 sentence :)

In current setup:
1. After following a citation, the entry is only pre-selected, and user needs to select it again to take any action.
2. Not all actions are applicable for citations (I wouldn't want to insert the whole bibtex entry into a citation)
3. I prefer to have helm-bibtex take the whole buffer, but selecting from a few follow actions doesn't need the whole buffer.

This PR solves 1 and adds better defaults and customization options for 2 and 3.

> And woudn't it be better to just replace the existing processor?

I'm not sure what you mean. This is exactly what this PR does!

tmalsburg:
Looks good. Could you please also update the documentation [here](https://github.com/tmalsburg/helm-bibtex#use-helm-bibtex-or-ivy-bibtex-as-an-org-cite-follow-processor) to reflect these changes?  Thank you!

mpedramfar:
Sure! I just made some changes. Let me know if there's anything else here I can change.

tmalsburg:
merged, thank you!

Issue #408, opened 2022-07-16
Feature request: Don't downcase titles


moritzschaefer:
While I see that it was a concious decision to downcase paper titles when formatting references, I am unable to grasp its benefit. 

As a hotfix, I just deleted the "downcase" in the following line:
https://github.com/tmalsburg/helm-bibtex/blob/ce8c17690ddad73d01531084b282f221f8eb6669/bibtex-completion.el#L1203

An elegant solution would be to introduce a customizable variable (helm-bibtex-keep-title-casing or something along the lines).

tmalsburg:
When displaying references this package loosely follows APA style.  I understand that APA is not relevant for everyone.  But it does make sense to follow some style.  IMO the correct solution for this is would be a configuration variable that specifies the style and then helm-bibtex should follow that style everywhere.  There should also be a plug-in architecture so that other people can add their own styles.   Configuration variables for controlling individual aspects of the presentation would be too fine-grained and lead to difficult-to-maintain code, I'm afraid.  I'm going to close this issue, but if you're interested in implementing the solution outlined above, please feel free to open a new issue for that.  I should say thought, that I'm happy with APA format and won't be able to invest any serious amount of time into implementing support for other styles.

PR #415, opened 2022-11-01
Implement custom variable to control reference formatting style

moritzschaefer:
This is a WIP PR implementing #408

It is intended as a minimal change to open the doors for potential subsequent style implementations. At the moment, the only style addition is to allow preserving title cases.

Feedback welcome, no thorough testing done yet

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -59,6 +59,11 @@
   "Helm plugin for searching entries in a BibTeX bibliography."
   :group 'completion)
 
+(defcustom bibtex-reference-style "apa"
+  "The reference style used when writing the reference."
+  :group 'bibtex-completion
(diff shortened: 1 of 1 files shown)


moritzschaefer:
I unfortunately don't have the time to propose anything more elaborate at the moment, but thought this would for the time being cover the two most important use-cases: 
1. Have a standardized lower-case title setting
2. Have a title case-preserving setting

Will have to see what time brings then.In any case, thank for your feedback and the generally awesome package.

tmalsburg:
Sorry that I couldn't give you a more positive response.  Unfortunately, this project has accumulated a bit of technical debt and maintainability is increasingly a concern. I we will find a good solution for this in the future.  For the time being I recommend that you simply define a custom variant of the relevant formatting function.

Issue #350, opened 2021-01-08
Control "do-not-find-pdf" argument with a variable


ashiklom:
For moderately large libraries (mine is ~3 MB) with lots of PDF files, the `bibtex-completion-find-pdf` call to `bibtex-completion-prepare-entry` becomes a significant bottleneck (see profile under "Details"). It would be nice to be able to disable it with a variable such as `bibtex-completion-do-not-find-pdf`. An easy modification would be to modify these lines:

https://github.com/tmalsburg/helm-bibtex/blob/master/bibtex-completion.el#L824-L826

...to something like this:

```
           (entry (if (and (not do-not-find-pdf) (not bibtex-completion-do-not-find-pdf) (bibtex-completion-find-pdf entry))
                      (cons (cons "=has-pdf=" bibtex-completion-pdf-symbol) entry)
                    entry))
```

The default for `bibtex-completion-do-not-find-pdf` would be `nil` to preserve existing behavior.

<details>

```
            - bibtex-completion-candidates                                                     1713  67%
             - apply                                                                           1713  67%
              - #<compiled -0xa9a41cb876ce6d0>                                                 1713  67%
               - bibtex-completion-parse-bibliography                                           957  37%
                - bibtex-completion-prepare-entry                                               786  30%
                 - bibtex-completion-find-pdf                                                   689  27%
                  - bibtex-completion-find-pdf-in-library                                       689  27%
                   - -first                                                                     681  26%
                      f-file?                                                                   681  26%
                   + bibtex-completion-get-value                                                  3   0%
                   + f-join                                                                       3   0%
                   + s-concat                                                                     2   0%
                 + mapcar                                                                        88   3%
                 + bibtex-completion-remove-duplicated-fields                                     8   0%
                   member-ignore-case                                                             1   0%
                + parsebib-read-entry                                                           168   6%
                  parsebib-find-next-item                                                         1   0%
               + insert-file-contents                                                           576  22%
```
</details>

tmalsburg:
[...]

Issue #226, opened 2018-01-27
Large difference in speed when using one notes file versus many notes files


jabranham:
I noticed that the first call to `helm-bibtex` was always quite slow and decided to investigate. Long story short, the call to `bibtex-completion-prepare-entry` was taking a long time. It turns out that this snippet (part of `bibtex-completion-prepare-entry`) was to blame:

```emacs-lisp
(entry (if (or
                       ;; One note file per entry:
                       (and bibtex-completion-notes-path
                            (f-directory? bibtex-completion-notes-path)
                            (f-file? (f-join bibtex-completion-notes-path
                                             (s-concat entry-key
                                                       bibtex-completion-notes-extension))))
                       ;; All notes in one file:
                       (and bibtex-completion-notes-path
                            (f-file? bibtex-completion-notes-path)
                            (with-current-buffer (find-file-noselect bibtex-completion-notes-path)
                              (save-excursion
                                (save-restriction
                                  (widen)
                                  (goto-char (point-min))
                                  (re-search-forward (format bibtex-completion-notes-key-pattern (regexp-quote entry-key)) nil t))))))
                      (cons (cons "=has-note=" bibtex-completion-notes-symbol) entry)
                    entry))
```

Since I had one large file, I hit the second part (all notes in one file). The regex search is slow, especially when it has to do this 1000 times to parse every entry in the `bib` file. I noticed that the other code looks like it would be faster and decided to convert my large org file into many small org files. There's not a built-in way to do this though, so I wrote the following function:

```emacs-lisp
(defun my/prop-to-file ()
  "Take the Custom_ID property and make a new file with just the entry"
  (let* ((props (org-entry-properties (point) "Custom_ID"))
         (id (cdr (assoc "CUSTOM_ID" props)))
         (filename (concat "/path/to/new/notes/directory/" id ".org")))
    (org-copy-subtree)
    (with-temp-file filename (yank))))
```

This assumes that your org file has note entries setup where the `Custom_ID` property is the bib key. I then opened up my bib file, widened the org buffer, put point on the first heading, and recorded a keyboard macro that applies this function then moved to the next heading. Then I applied that macro to all the headings in the buffer (which took a while).

After changing the relevant variables (like `bibtex-completion-notes-path`), the first call to `helm-bibtex` is about 5 or 7 times faster. It's quite noticible. 

Dropping this note here with the hope that maybe it helps someone else.

tmalsburg:
[...]

PR #419, opened 2022-12-18
optimize bibtex-completion-prepare-entry

mpedramfar:
If there are a lot of folders to search through, especially if some of them are on the cloud, `bibtex-completion-find-pdf` can be slow.
Currently, `bibtex-completion-prepare-entry` searches for the pdf files for all the entries.
If `bibtex-completion-pdf-symbol` is the empty string, then it will still search for every single pdf file and then show nothing for it.

This pull request checks if `bibtex-completion-pdf-symbol` is the empty string, and if it is, it will skip searching for it.
The same is done for notes and `bibtex-completion-notes-symbol`.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -107,7 +107,7 @@ having a PDF if \"<key>.<extension\" exists."
 
 (defcustom bibtex-completion-pdf-symbol "⌘"
   "Symbol used to indicate that a PDF file is available for a publication.
-This should be a single character."
+This should be a single character or the empty string."
   :group 'bibtex-completion
   :type 'string)
 
@@ -181,7 +181,7 @@ directory (not a file)."
 
 (defcustom bibtex-completion-notes-symbol "✎"
   "Symbol used to indicate that a publication has notes.
-This should be a single character."
+This should be a single character or the empty string."
   :group 'bibtex-completion
   :type 'string)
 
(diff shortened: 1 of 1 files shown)


tmalsburg:
I think you don't want to show PDF symbols in the results list but still be able to open PDFs? If correct, I see how this PR fixes your performance issue, but it seems like an extension that is very specific to your personal setup.  I also don't understand why finding PDFs is so slow in your setup.  Even with many folders it should be fast. I suspect that the solution for this issue may lie somewhere else. Question: What do you mean when you say that some folders are in the cloud?  Are these Dropbox folders or something like that?  These should be cached locally and therefore shouldn't be slow.

mpedramfar:
Yes. I don't want to lose any functionality and I don't want any changes in the behavior (even when `bibtex-completion-pdf-symbol` is the empty string) except for speed.

It probably is a bit specific to my setup. However, if you use rclone to mount a cloud storage (Dropbox, Google Drive, One Drive, etc) or use sshfs to mount a remote folder, then even running `ls` to list the contents of a remote directory could be slow.

mpedramfar:
This pull request also resolve #350 and provides a workaround for #226

tmalsburg:
Not sure but I think a better solution might be to read the list of available PDFs once and then to check that list, not each file separately on disk.  In general, we need a better approach to detecting PDFs and I have an idea for a plug-in approach that users could flexibly configure to their specific needs.  Such a system would close half of all issues for this project or so.  Perhaps I will find some time to work on this over the holidays but I can't promise it.

tmalsburg:
How many folders to you have? Asking because it's useful for me to know what setups users have.

mpedramfar:
I have less than 500 folders, but some of them are in the cloud (mounted network files, not cached) and I've also made some tweaks to other places in bibtex-completion so that my pdf files are not necessarily of the format citekey.*.pdf
Because of this, the search becomes very slow. 
So, in a way, this PR helps with 2 issues for me:
1. slow access to network mounted drive
2. support for more sophisticated pdf file names (which inevitably slows down the search, no matter how it's implemented)

On the other hand, I'm not entirely sure how a plug-in approach could be different enough to not have the first issue.
As I understand, if the new system is at least as capable as the current one, then this PR can help.

By capable, I mean this:
**Are we caching a list of pdf files, or are we detecting them?**
If we are caching the list, then we need to watch for any change in the filesystem. In linux, except for local directories, this can be a bit of a hit and miss (e.g. see https://github.com/libfuse/libfuse/wiki/Fsnotify-and-FUSE and https://github.com/fsnotify/fsnotify#why-dont-notifications-work-with-nfs-smb-fuse-proc-or-sys). So, unless we wait until inotify support has been added for various FUSE and network filesystems, this is a loss in capability.
One way to "watch" for changes is to periodically detect them, which brings us back to the issue of detecting.
If we're going to detect them, then the cost of accessing the filessystem is always there, especially in network mounted folders. And this is exactly where this PR helps.

On the other hand, I can't think of how this change could be bad. Do you have a specific situation/plug-in system in mind where this change can be undesirable?

PR #435, opened 2023-07-14
Improve formatting of APA-style citations when biblatex files are used

swflint:
Currently, only the field `journal` is pulled from when formatting APA-style citations.  Users of biblatex may store journal name information in `journaltitle` instead.  This allows for either name to be used in templates, and gets the data from whichever field is filled, `journal` preferred.


tmalsburg:
Merged. Thank you.

PR #439, opened 2023-09-17
fix year extraction from date in bibtex-completion-apa-get-value

daniron26:
line 1223 is not reachable:

(car (split-string (bibtex-completion-get-value "date" entry "") "-"))))

it is active when "year" is nil, but enclosed in a an if clause where "year" in non-nil. This prevent year from being extracted from date in case where there is only date and not year.


tmalsburg:
Merged, thank you.

PR #442, opened 2023-11-21
enable evaluation of date field if year is not defined

dominikpegler:
Without this addition to the if clause, in cases where the `year` field is missing, the line where `date` substitutes `year` will never be evaluated. This means that the year is missing when inserting the reference into an org-file, but also in org-roam-ui, for example (e.g. "Author (). Title").

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -1191,7 +1191,7 @@ string if FIELD is not present in ENTRY and DEFAULT is nil."
      (_
       ;; Real fields:
       (let ((value (bibtex-completion-get-value field entry)))
-        (if value
+	(if (or value (and field (string= field "year")))
             (pcase field
               ;; https://owl.english.purdue.edu/owl/resource/560/06/
               ("author" (bibtex-completion-apa-format-authors value))


tmalsburg:
I don't understand what problem this PR is trying to fix.  This function seems to work as intended.  If `value` is `nil` the second `pcase` on line [1224](https://github.com/tmalsburg/helm-bibtex/blob/master/bibtex-completion.el#L1224) is evaluated.  Just tried it with this entry

```
@Article{test,
  author    = {Chater, Nick and Oaksford, Mike},
  title     = {Ten years of the rational analysis of cognition},
  journal   = {Trends in Cognitive Sciences},
  Date      = {1998-6-15},
  volume    = 3,
  number    = 2,
  pages     = {57–65},
  issn      = {1364-6613},
  doi       = {10.1016/s1364-6613(98)01273-x},
  publisher = {Elsevier BV},
}
```
and this code

``` elisp
(let ((entry (car (bibtex-completion-candidates))))
  (bibtex-completion-apa-get-value "year" entry))
```
This returns 1998, as expected.

tmalsburg:
Note that this is not the repository of org-roam.  Perhaps you're running into a bug there.

Issue #297, opened 2019-07-05
Add a configuration option to prevent proactive reloading


mwlodarczak:
While I understand its rationale, I find the new proactive reloading of bibliography quite frustrating as it kicks in every time I save the BIB file. I wonder whether adding a variable that controls this behaviour would be an option?

tmalsburg:
While I sympathize, I'm worried about inflation of customization variables; we already have too many if you ask me.  Perhaps a better solution would be to reload the bibliography in the background so that users don't have to wait?  This could be implemented using threads which were introduced in Emacs 26.

mwlodarczak:
Titus von der Malsburg <notifications@github.com> writes:

> While I sympathize, I'm worried about inflation of customization
> variables; we already have too many if you ask me. Perhaps a better
> solution would be to reload the bibliography in the background so that
> users don't have to wait? This could be implemented using threads
> which were introduced in Emacs 26.

Indeed, that would be a good compromise. I understand the desire to keep
the number of customisation variables to a minimum but I would argue
that reloading on each save is such an invasive behaviour that it does
warrant making an exception.

Marcin

tmalsburg:
Could you make a PR please?

mwlodarczak:
Sure. I think that's all it takes: [PR #301](https://github.com/tmalsburg/helm-bibtex/pull/301).

rdiaz02:
If I may, I'd like to raise this issue again. I've fixed what I think was a bug in the original PR and added a longer docstring that specifically addresses @tmalsburg question:

> If someone loads helm-bibtex with the default setting for bibtex-completion-watch-bibliography and then decides that they don't want to proactively reload the bibliography, they have to restart Emacs with the new setting, right?

The answer is yes: changing the value of the variable after loading helm-bibtex has no effect.

This is the PR: https://github.com/tmalsburg/helm-bibtex/pull/443

tmalsburg:
See `bibtex-completion-watch-bibliography`.

PR #443, opened 2024-01-06
Add a customisation for toggling of proactive reloading of bibliography, II.

rdiaz02:
From the change of mwlodarczak at

https://github.com/mwlodarczak/helm-bibtex/commit/4a421cae9b7d4cdb4a0933080633564b1774addb

with a minor fix and more detailed docstring.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -77,6 +77,20 @@ composed of the BibTeX-key plus a \".pdf\" suffix."
   :group 'bibtex-completion
   :type '(choice directory (repeat directory)))
 
+;; From https://github.com/mwlodarczak/helm-bibtex/commit/4a421cae9b7d4cdb4a0933080633564b1774addb
+(defcustom bibtex-completion-watch-bibliography t
+  "If non-nil (the default) the bibliography is reloaded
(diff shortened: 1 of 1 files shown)


tmalsburg:
Thank you for making this PR.  Could you please provide some context? I'm wondering why anyone would not want to reload the bibliography when it was changed?

tmalsburg:
Ah, I just saw your comment in #297.  Hm, I think the solution to reload asynchronously (in a thread) would be much nicer. Doesn't need a configuration variable and should work for everyone.  Unfortunately, I don't have to time to implement it.  There is at least one possible race condition that would have to be dealt with carefully: a second reload is triggered before the first is finished. What do you think?

rdiaz02:
I think you are absolutely right that to reload asynchronously (in a thread) would be much nicer. But my elisp is completely rudimentary, so I would not be able to implement it.

P.S. Apologies for not having provided context in the PR.

tmalsburg:
Thanks for the response.  I merged the PR as a stop-gap solution.  Hope we will have something better in place in the future.

rdiaz02:
Thank you.

PR #446, opened 2024-02-20
Support AUCTeX's (new) LaTeX-mode

dankessler:
AUCTeX (14.0.1+) has renamed latex-mode -> LaTeX-mode. As a result, inserting citations in buffers that are in `LaTeX-mode` result in `bibtex-completion-insert-citation` falling back on the default `bibtex-completion-format-citation-default`, which just inserts the naked key without wrapping it in a `\cite` command (or similar).

A simple hack is to explicitly support both `latex-mode` *and* `LaTeX-mode`


tmalsburg:
Merged, thank you.

Issue #452, opened 2024-11-04
New `parsebib.el` parser


joostkremers:
Hi @tmalsburg , `bibtex-completion.el` depends on `parsebib.el`, so I wanted to give you a heads-up: I've ripped out the pretty hairy regexp-based parser code and replaced it with a simple recursive descent parser. The new parser works well, and the tests included in `parsebib` all pass, but with this kind of thing there's always a risk of regressions, of course.

I haven't merged the new code yet, but once I do, you may get bug reports from users that are ultimately caused by `parsebib.el`. If that happens, just send them to me, or key me in in the discussion.

Some functions have also changed signature, especially `parsebib-read-entry`, which is used in `bibtex-completion.el`. You may want to consider using `parsebib-parse-bib-buffer` or `parsebib-collect-bib-entries` instead, which will get you all entries in a single function call.

If you have a `.bib` file that you would like to test the new code with before I do the merge, I'd be happy to do that, or if you prefer, you can check out the code [here](https://github.com/joostkremers/parsebib/tree/wip/rdp).

chuxubank:
I just find a bug that `bibtex-completion-parse-strings` runs in infinite loop when use with new `parsebib` package.
I pin the `parsebib` to the latest release version (4.7) and it still works well.

So if we want fix it, we should do the change on `bibtex-completion` side, right?

joostkremers:
> So if we want fix it, we should do the change on `bibtex-completion` side, right?

Depends. `bibtex-completion` needs an update in order to be compatible with the new `parsebib.el`. I submitted a PR, but it looks like it hasn't been merged yet. Keeping `parsebib` pinned on 4.7 is the best solution for now. I'm sure @tmalsburg will update `bibtex-completion` soon.

tmalsburg:
Just merged the PR.  Thank you Joost!!

lnguyen4:
Hi, I m having the issue that when I invoke `helm-bibtex` , it hangs and the mini-buffer says "Parsing bibliography file /home/directorty/myfile.bib". 

I got this issue last night and came here to notice there was an update on the dependency `parsebib` pacakge. So, I then downgraded the parsebib and the issue was gone. 

[...]

PR #453, opened 2024-11-05
Update bibtex-completion.el for parsebib.el 6.0

joostkremers:
As I mentioned in issue #452 , I've made some changes to the `parsebib.el` API that affect `bibtex-completion.el`. This PR fixes those issues. Obviously, it shouldn't be merged until I push the new `parsebib.el` to master.

I've only changed the relevant function calls. Given the purpose of `bibtex-completion.el`, it would probably make sense to use `parsebib-collect-bib-entries` and `parsebib-collect-strings`, but that would require a bigger change (e.g., `parsebib-collect-bib-entries` can do some of the things that `bibtex-completion-prepare-entry` does, it seems), so you probably have a better idea if that's worth it.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -5,7 +5,7 @@
 ;; Maintainer: Titus von der Malsburg <malsburg@posteo.de>
 ;; URL: https://github.com/tmalsburg/helm-bibtex
 ;; Version: 1.0.0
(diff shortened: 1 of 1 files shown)


tmalsburg:
Thank you for this PR.  I'll merge it when you give me the green light.

joostkremers:
Thanks. I'd like to merge the new parser some time over the weekend. I'll let you know when I do.

tmalsburg:
Merged.  Thank you!!

PR #454, opened 2024-11-14
Avoid infinite loop when parsing a .bib file.

joostkremers:
In parsebib.el 4.7 / 5.0, the function parsebib-find-next-item leaves point after the @-sign when it finds an item. In parsebib.el 6.0, it leaves point *before* the @-sign. Therefore, the caller must advance point if the item is not read, otherwise parsebib-find-next-item will keep finding the same item.

This fixes the problems reported in https://github.com/tmalsburg/helm-bibtex/issues/452 and https://github.com/jkitchin/scimax/issues/510.

### DIFF
--- a/bibtex-completion.el
+++ b/bibtex-completion.el
@@ -496,7 +496,8 @@ for string replacement."
                    if (string= (downcase entry-type) "string")
                    collect (let ((entry (parsebib-read-string ht)))
                              (puthash (car entry) (cdr entry) ht)
-                             entry))))
+                             entry)
+                   else do (forward-line 1))))
     (-filter (lambda (x) x) strings)))
 
 (defun bibtex-completion-update-strings-ht (ht strings)
@@ -684,7 +685,7 @@ If HT-STRINGS is provided it is assumed to be a hash table."
                                bibtex-completion-additional-search-fields))
    for entry-type = (parsebib-find-next-item)
    while entry-type
-   unless (member-ignore-case entry-type '("preamble" "string" "comment"))
+   if (not (member-ignore-case entry-type '("preamble" "string" "comment")))
    collect (let* ((entry (parsebib-read-entry nil ht-strings))
                   (fields (append
                            (list (if (assoc-string "author" entry 'case-fold)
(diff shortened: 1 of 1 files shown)


tmalsburg:
Thank you so much, Joost 🙏

uliw:
OK, so in the meantime, I replaced helm with ivy, and I downloaded the new bibtex-completion.el. Noe, when I do `org-ref-insert-cite-link` I get the following backtrace
```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  member-ignore-case(nil ("comment" "preamble" "string"))
  parsebib-read-entry(nil #<hash-table equal 0/65 0x156c699342d1>)
  bibtex-completion-parse-bibliography(#<hash-table equal 0/65 0x156c699342d1>)
  bibtex-completion-candidates()
  org-ref-read-key()
  org-ref-insert-cite-link(nil)
  funcall-interactively(org-ref-insert-cite-link nil)
  command-execute(org-ref-insert-cite-link record)
  counsel-M-x-action("org-ref-insert-cite-link")
  ivy-call()
  ivy-read("M-x " ("toggle-debug-on-error" "package-recompile-all" "package-list-packages" "imenu" "org-modern-mode" "org-ref" "org-ref-insert-link" "org-ref-bibtex-hydra/body" "ert" "org-ref-bibtex-hydra/nil" "ag" "cd" "5x5" "arp" "dbx" "dig" "erc" "eww" "ftp" "gdb" "irc" "jdb" "man" "mpc" "ore" "pdb" "pwd" "rsh" "sdb" "wos" "xdb" "bash" "calc" "diff" "dirs" "esup" "ffap" "gnus" "grep" "help" "ielm" "info" "life" "mail" "mpuz" "navy" "ping" "pong" "smex" "talk" ...) :predicate counsel--M-x-externs-predicate :require-match t :history counsel-M-x-history :action counsel-M-x-action :keymap (keymap (67108908 . counsel--info-lookup-symbol) (67108910 . counsel-find-symbol)) :initial-input nil :caller counsel-M-x)
  counsel-M-x()
  funcall-interactively(counsel-M-x)
  command-execute(counsel-M-x)
```

Is this unrelated? BTW, I noticed that org-ref retains a dependency on helm-bibtex, even when I use ivy instead of helm

joostkremers:
Oh boy, I just wrote somewhere else that I was hoping I fixed all issues, but this seem to be another one... :sob: 

Though honestly, this is a bit of a riddle to me. I can read your bibliography just fine using `bibtex-completion-candidates` or `bibtex-completion-parse-bibliography`, and I don't see how a call to `parsebib-read-entry` is *followed* by a call to `member-ignore-case`...

uliw:
once the above fixes are on elpa, I'll do a reinstall and test with a minimal startup file.

PR #455, opened 2024-11-15
(bibtex-completion-get-entry1): Position point correctly.

joostkremers:
Since parsebib 6.0, `parsebib-read-entry` expects point to be before the entry to be read. 

Also remove the let-binding for `entry-type`, since it's no longer needed.


tmalsburg:
Merged, thank you.

